PR #182, opened 2014-12-15
Autogenerated publication list from BibTeX file

jiahao:
This PR adds a bunch of papers from the recent HPTCDL 2014 workshop featuring Julia.

Additionally, there is a revamp of how the publication list works:
- The publications are now specified in a BibTeX file, `julia/publications.bib`
- The publications are categorized in the listing in `publications/_EDIT_ME_index.md`
- The final publication list in `publications/index.md` is now autogenerated from the two files above with  `make -C publications`.

Maintaining a single BibTeX file and CSL file for citation formatting considerably simplifies the management of the citation list going forward, as the build process automatically numbers the citations while allowing future flexibility in formatting the citations consistently, and furthermore allows users to download the references in a standard format.

Technical note: The build mostly uses `pandoc` and `pandoc-citeproc` to generate the citations. Unfortunately the version of `pandoc` I have (v1.13.1) does not support automatic citation generation into Markdown format, so one has to resort to some considerably ugly hacks to splice the Jekyll header and HTML-formatted citations together. Furthermore I have not been able to make `pandoc-citeproc` spit out enough vertical space to align the abstracts nicely, so that is also hacked in using `sed`. I've also had to edit out LaTeX math environments in abstracts, as `pandoc-citeproc` automatically inserts vertical whitespace at the end of every `$...$` environment, and inconsistently strips certain LaTeX commands (like \enumerate) while silently passing others (like {\rm ...}); the latter causes Jekyll to choke.


carlobaldassi:
I hadn't looked at the publications section in a while. Is the penultimate section intended for papers which make use of Julia code, or is there some other requirement? I have [one such paper](http://www.plosone.org/article/info%3Adoi%2F10.1371%2Fjournal.pone.0092721), and the [corresponding code](https://github.com/carlobaldassi/GaussDCA.jl) (we actually provide both matlab and julia code, but the julia one is slightly faster and doesn't use C, while the matlab one needs those horrible mexfiles...).
Should I add it?

jiahao:
Yes, please do! I added all the papers I could find citing the 2012 preprint on Google Scholar.

carlobaldassi:
ok will do! (we cited the website...)

StefanKarpinski:
Should we do a PSA on julia-users asking for people to submit PRs for papers they've published that have Julia code?

jiahao:
Sure, why not.

ViralBShah:
Good idea. Also, I wonder if some JS wizardry can be done so that the abstracts can be expanded, but otherwise you just get a list of publications, rather than a wall of text.

PR #188, opened 2015-01-03
Add packages link, remove rss link

### LINKED ISSUE #112: Packages link in links bar?
I'd submit a PR but I'm not really sure where it should go

Could link to http://iainnz.github.io/packages.julialang.org/ until DNS stuff gets sorted

jiahao:
Closes #112.

@simonbyrne's suggestion of removing the rss link provides just enough space so that "packages" can be squeezed in.

'Learning', 'teaching' and 'publications' could probably be consolidated in a future change.

screen shot 2015-01-03 at 12 37 21 am

### DIFF
--- a/_includes/links.html
+++ b/_includes/links.html
@@ -4,12 +4,12 @@
     <a href="https://github.com/JuliaLang/julia">source</a>&nbsp;|
     <a href="/downloads/">downloads</a>&nbsp;|
     <a href="http://docs.julialang.org/">docs</a>&nbsp;|
+    <a href="http://pkg.julialang.org/">packages</a>&nbsp;|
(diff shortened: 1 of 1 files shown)


ViralBShah:
+1

Learning and teaching can be merged as well, but publications should probably remain separate.

IainNZ:
Will be interesting to see if there is bump in analytics, hovering at about ~900 "sessions" per week over last 14 days

ihnorton:
The julia link in the bar is redundant, because the logo points to the same link.
I think we could also take out the juliacon link and add it to community. Once this year's date is announced, we could put it back, or preferably put it next to the logo for prominence, e.g.:

image

(a little bit nicer than that, of course)

PR #189, opened 2015-01-04
Juno download links

one-more-minute:
@ViralBShah suggested adding some Juno download links to the Julia downloads page, so I've tried to do that in a sane way. Suggestions are very welcome.

Incidentally, this page could probably be earmarked for a rewrite, a lot of it is crufty or even out of date.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -40,6 +40,35 @@ We provide stable Julia packages and nightly releases on various operating syste
 </tr>
 </tbody></table>
 
+## Julia + Juno IDE bundles
+
+Julia works very well when used through the Terminal or Command Prompt interface,
+but if you're looking for more IDE-like features (such as integrated code editing
+and evaluation) we recommend [Juno](http://junolab.org). On Windows and OS X we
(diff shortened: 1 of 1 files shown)


ViralBShah:
Should we put some text and a link to Julia-LT there about where to file Juno related issues?

ViralBShah:
Are these codesigned? It didn't alert me on the mac, but that may be perhaps because I had given permissions to the OS to allow Juno, when I first tried it out.

ViralBShah:
The downloads page rewrite has been on my mind for a while. To start with, we can just move much of the OS level details and graphics to a different page, and make this simply about getting to Julia: REPL-based, Juno, or JuliaBox. I think it will be nice to add some screenshots too.

one-more-minute:
+:100: to some Julia/JuliaBox/Juno screenshots, that could really help draw people in. Might be nice to have those on the home page along with a nice big "download" link.

The docs explain where to get help and post issues, but it is a little indirect; I'll add more info on help.

And yes, builds are now all codesigned thanks to @staticfloat :)

PR #193, opened 2015-01-08
Update download links for v0.3.5

staticfloat:
Waiting for binaries to actually be uploaded

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -3,7 +3,7 @@ layout: default
 title:  Julia Downloads
 ---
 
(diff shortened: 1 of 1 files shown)


staticfloat:
Alright, looks like all the binaries are reachable.  Launchpad is crunching together the binaries for the PPA, and we're waiting on @nalimilan to build the `.rpm`'s, but other than that I believe we are good to go.

tkelman:
May as well make status.julialang.org/stable point to 0.3-latest on S3 so they don't have to be updated separately any more?

PR #199, opened 2015-01-20
Use HTTPS URL for sites that support it

scw:
Juliabox, S3, and status.julialang.org URLs all support HTTPS; use these as the default links.


ivarne:
CC @staticfloat

staticfloat:
Yes, thank you very much for making sure these are consistent.

PR #203, opened 2015-02-01
Update the link and description for the JuliaQuantum org.

i2000s:


i2000s:
Now it should be in a good shape. Thanks.

PR #206, opened 2015-02-11
add "Quantitative Economics with Julia" to Topical

catawbasam:


StefanKarpinski:
Thanks!

ViralBShah:
We should probably also link to their Julia tutorial as the first thing in `teaching`.

Cc: @amitmurthy

PR #208, opened 2015-02-14
Minor corrections for UNAM courses; added Luis Benet's course in Fall 20...

dpsanders:
...14

cc @lbenet


jiahao:
Ah, capitalization.

PR #213, opened 2015-02-19
Add a Julia Bignum GSOC project

wbhart:
As requested, here is an outline of a GSoC project for improving Julia Bignums.


ViralBShah:
Thanks.

PR #214, opened 2015-02-20
Update index.md

TPieciak:
We use Julia language for signal processing in medical diagnostic systems course in Poland.


ViralBShah:
Awesome. Do you use any particular packages? Just curious.

TPieciak:
Mostly, we are going to prepare our own packages for biomedical signal
processing, in particular electrocardiography. I am also using Julia for
magnetic resonance imaging for my own research, i.e., implementation of the
recent work:
http://www.sciencedirect.com/science/article/pii/S1361841514001625

Julia becomes everywhere!

ViralBShah:
That's awesome. It would be great if you can submit a PR linking your paper to the publications section also, if not already there.

PR #216, opened 2015-03-11
Update index.md

Ivo-Balbaert:


StefanKarpinski:
Exciting that there's an actual Julia book out there now!

PR #221, opened 2015-03-27
Update benchmarks

jiahao:
The main reason for this update is to represent Fortran's performance more accurately, due to a change in how we benchmarked the Fortran code.

Also bumps some versions of various software being compared.

Ref: JuliaLang/julia#10157


ViralBShah:
Thanks for doing this.

PR #227, opened 2015-05-01
Update download links for v0.3.8

tkelman:


staticfloat:
Looks like we're good to go here.  Nice work, @tkelman and co.!

PR #228, opened 2015-05-08
community: Update mailing list section

jayschwa:
* Added julia-news
* Changed URLs to a slightly shorter, nicer form
* Changed from paragraphs to an unordered list
* DRY'd out some of the wording and other small tweaks

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -10,17 +10,13 @@ them in all Julia related forums.
 
 # Mailing Lists
 
-The [julia-users][julia-users] mailing list is for discussion around the usage of julia. New users of julia can ask their questions here. As a courtesy to others, do check the archives and documentation to see if a particular question is already answered.
-
-The [julia-dev](https://groups.google.com/forum/?fromgroups=#!forum/julia-dev) mailing list is for discussion related to the development of julia itself.
-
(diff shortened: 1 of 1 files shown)


jayschwa:
I suggest locking down julia-news and making it read-only before merging this. It appears anybody can post to the list right now.

StefanKarpinski:
That was the intention but I guess that replies could go to Julia-users or something like that.

pao:
I think the intent with -news is that if it's not a Julia (pre-)release announcement, it doesn't belong there--sysadmins or folks who play them in their schools or offices would be likely to need to act on something posted there.

jayschwa:
Right. There are plenty of other lists for discussions. This shouldn't be one of them. The model I had in mind was https://groups.google.com/group/golang-announce.

StefanKarpinski:
I've changed it so that only managers and owners can post. I also made you a manager, Jay :-)

jiahao:
Should the list be called `julia-announce` instead? I feel like I've seen `-announce` more frequently than `-news`.

StefanKarpinski:
Already done and I don't really feel like making a new list.

jayschwa:
I think I've seen "announcement" more too, but I doubt "news" will confuse anyone.

ViralBShah:
Yes, `-announce` is more common, but no point creating a new one now. `-news` is perfectly fine.

jayschwa:
Bump. We had our inaugural julia-news post, but the list still isn't mentioned on the site as far as I can tell.

PR #231, opened 2015-05-12
JuliaCon banner

IainNZ:
Add a prominent, temporary JuliaCon banner to the main page. I think given we are 1.5 months out, this makes sense to really advertise. This is what it looks like:


IainNZ:
Just need one person to tell me this isn't stupid and I'll merge. @jiahao, can you be that person? :)

jiahao:
get-the-party-started1

PR #234, opened 2015-05-23
RFC: draft of jsoc announcement

mlubin:
As requested by @ViralBShah.

Should we set up an email group (a la juliacon) to manage inquiries and submissions?

### DIFF
--- a/_includes/links.html
+++ b/_includes/links.html
@@ -10,6 +10,7 @@
     <a href="/learning/">learning</a>&nbsp;|
     <a href="/teaching/">teaching</a>&nbsp;|
     <a href="/publications/">publications</a>&nbsp;|
+    <a href="/blog/2015/05/jsoc-cfp/">jsoc</a>&nbsp;|
     <a href="http://juliacon.org">juliacon</a>
   </p>
   <h1 style="border-top: 1px solid #ddd;">JuliaCon 2015</h1>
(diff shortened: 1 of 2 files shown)


ViralBShah:
+1 on a juliasoc mailing list!

mlubin:
Just thinking now, should this be a private list for organizers or a public list for everyone?

ViralBShah:
For experienced Julia contributors, we can still have them in JSOC, but they would still need a mentor - just as someone to approve that the work was done. Do we need to mention this separately?

mlubin:
Yes, that's here:
> If you're already a contributor to Julia or a Julia package and want to get paid to continue an existing project, that's okay too! In this case we still ask you to find a mentor who's familiar with your field of work.

mlubin:
We can tweak the wording of course

ViralBShah:
Thanks. The wording is fine.

Actually, we also do not need to restrict to students. All are welcome, so long as they can spend a good chunk of time doing the projects that they submit - with preference for students.

mlubin:
Are there any official qualifications then? I'm assuming we can't fund people based in Cuba or Iran

ViralBShah:
Not at the moment. We will have to check with NF. Do you know what GSOC says on this?

ViralBShah:
I think we should put something generic here about selections being made in conformance to applicable Government laws.

What do you think about also adding `jsoc` to the menu items on the top, and also an advertisement next to the juliacon one?

mlubin:
Updated with menu item and working email link. Should be mergeable.
I'll let someone with better web skills set up an add next to the juliacon link.

PR #245, opened 2015-06-30
Remove JuliaCon 2015 banner

jiahao:
It was good while it lasted.


ihnorton:
We should replace it with Julia QOTD from @kshyatt's twitter feed!

jiahao:
The quote from @ViralBShah was my favorite: 'If you can avoid using sparse matrices, please don't use them!'

kshyatt:
I'm always in favour of putting more of my bad posting on the JuliaLang frontpage :wink:

PR #251, opened 2015-07-15
Añadir script de Transifex.

Ismael-VC:


Ismael-VC:
Sorry this was supposed to be for the Spanish version!

Issue #63, opened 2013-11-11
homepage revamp


StefanKarpinski:
Kicked off by this post: https://groups.google.com/forum/#!topic/julia-dev/pQG6bSglkjc, maybe the home page should get a redesign. This issue is a location for discussion and brainstorming ideas for such a redesign.

dpsanders:
Bump. Note that the parallel code
```
nheads = @parallel (+) for i=1:100000000
  int(randbool())
end
```
now actually has two bugs in 0.5!

ViralBShah:
Fixed that bug.

waldyrious:
> Fixed that bug - rather both.

a387af7 for those wondering.

ChrisRackauckas:
Bumping this. Whoever did your JuliaComputing design was really good. You should try to get them to redesign the Julia page as well!

StefanKarpinski:
We're planning on it. The content needs a lot of reorganization as well.

ViralBShah:
@shashi Can you post your redesign of the julialang front page. We can do the design work separately, but I think shashi got off to a good start.

shashi:
sure, it's at http://shashi.github.io/julialang.github.com/

you have to manually rewrite links for now (e.g. Learning is at http://shashi.github.io/julialang.github.com/learning/ and not http://shashi.github.io/learning/)

If someone offers content for the features section of the front page, I will put that in. Do mock up the layout you want to show it with, I can also put that in. All the internal pages basically look as they did before, but in better fonts and stylesheet. Just the new front page needs content.

ViralBShah:
We probably should do this for 1.0

shashi:
Anyone volunteering to write content for the home page? I'd like to keep the click on feature -> short description + example layout.

ViralBShah:
I wonder if we should close this issue, and have a slack #website channel. I wonder if @ChrisRackauckas has some time to brainstorm this. While I think the prototype website has good content, it is not user focussed. It does not convince someone coming why they should use Julia. It may be better to showcase Julia packages front and center along with the language on the landing page, for example.

ChrisRackauckas:
## The Constraints

I think this is a difficult issue because the webpage has to stand somewhere in the battle of Julia's identity between a general-purpose programming language and a tool for technical computing. Right now it's an odd hybrid. For example, if you take a look at Rust's homepage:

https://www.rust-lang.org/en-US/

[...]

PR #253, opened 2015-07-22
Justify paragraphs.

Ismael-VC:
Justified alignment looks better IMHO.

After: 

captura de pantalla 2015-07-22 a las 15 34 52

Before:

captura de pantalla 2015-07-22 a las 15 35 37

Demo: http://julialanges.github.io

### DIFF
--- a/css/screen.css
+++ b/css/screen.css
@@ -65,6 +65,7 @@ h2 {
 
 p {
   margin: 1em 0;
+  text-align: justify;
 }
 
 a {


pao:
This also uncenters the navigation bar. I don't think that was intended.

That side, I'm not convinced I like this any better. It makes the wall of text even more wall-like. But y'know, opinions.

jiahao:
-1

Most typographers recommend against full justification of text. [This blog post](https://kaiweber.wordpress.com/2010/05/31/ragged-right-or-justified-alignment/) has a nice summary of the reasons why you shouldn't, as well as studies on the impact of full justification on decreased readability.

Furthermore, the W3C explicitly [recommends against full justification](http://www.w3.org/TR/WCAG20-TECHS/G169.html), stating:

> Many people with cognitive disabilities have a great deal of trouble with blocks of text that are justified (aligned to both the left and the right margins). The spaces between words create "rivers of white" running down the page, which can make the text difficult for some people to read. This failure describes situations where this confusing text layout occurs. The best way to avoid this problem is not to create text layout that is fully justified.

johnmyleswhite:
Jiahao, I think you've got this backwards. I believe the PR tries to remove full justification.

jiahao:
Derp. Thanks for spotting that.

pao:
Er, what? The current text is not justified, it's ragged-right. The style added in the diff is `text-align: justify`. What am I missing?

johnmyleswhite:
Is the original description wrong when showing before and after images?

jiahao:
Yes the original description is wrong. I need caffeine.

pao:
Ah, yeah, I didn't actually read the labels. They're wrong though so that didn't cost me anything.

Ismael-VC:
Sorry the description is wrong, navigation bar can be centered like this:

* https://github.com/JuliaLangEs/julialanges.github.com/blob/master/_includes/links.html#L2

captura de pantalla 2015-07-22 a las 15 51 41


I think not justified text looks messy.

Ismael-VC:
Oh the PR is closed, 

@jiahao The Julia web site is very simple, perhaps if it had more elements I would agree with you, but as it is right now it´s just the text and the white background, no visual style at all. 

I understand the "white river issue", but did you check the demo? I don´t see any unless I use a smartphone or a lot of zoom, but one way or another it looks ugly on a smartphone.

Anyway I'm glad accessibility is taken into account.

jiahao:
Yes I did, and I did verify that you can quite easily find a browser window size that exhibits this 'river' illusion.

now:

screen shot 2015-07-22 at 5 29 37 pm

this pr:

screen shot 2015-07-22 at 5 29 27 pm

Ismael-VC:
Ok, I see. Didn't consider that too much  I think that the thing that bothers me is that the web is too simple, do you know if there have been plans to redesign the web site?

I only found this: https://groups.google.com/forum/#!topic/julia-users/B5d-Wtp17GA

But they discuss content (which I know is more important) only.

jiahao:
You can submit a PR and have people look at it. But a redesign has already been tried - see #178, which was ultimately reverted.

Ismael-VC:
@jiahao thanks!

jiahao:
@Ismael-VC see also #63

PR #255, opened 2015-07-27
Update download links for v0.3.11

tkelman:
cc @staticfloat, I sorted out the buildbots and they're working on these now, once they're uploaded to julianightlies can you move over to the julialang bucket and merge? Thanks.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -3,7 +3,7 @@ layout: default
 title:  Julia Downloads
 ---
 
-# Current Release (v0.3.10)
+# Current Release (v0.3.11)
 
 We provide three ways for you to run Julia:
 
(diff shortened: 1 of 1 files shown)


staticfloat:
Oh good; any idea what was weird about them?

staticfloat:
Moved over and checksums uploaded.

tkelman:
Thanks! It turns out if you do something to the osx buildbots that causes gfortran to not be available, the "quickbuild" runs won't work correctly until you run at least one "package" build.

staticfloat:
That is the kind of errata I wish we didn't have to know. :P

tkelman:
That should be fixable by adding a few of the brew lines to the quickbuild steps, but they'd need to be conditional on darwin.

PR #256, opened 2015-08-07
Includes references to Learn Julia

chrisvoncsefalvay:
_Learn Julia_ is now in Early Access, and as such, I have included it in the list of learning resources.

### DIFF
--- a/learning/index.md
+++ b/learning/index.md
@@ -8,6 +8,11 @@ title:  Learning Julia
 - [Seven More Languages in Seven Weeks](https://pragprog.com/book/7lang/seven-more-languages-in-seven-weeks) by [Bruce Tate, Fred Daoud, Jack Moffit and Ian Dees] - published by The Pragmatic Programmers (350 pages, Published: 2014-11-15, ISBN: 978-1-94122-215-7)
     This book contains a Julia tutorial Chapter (written by Jack Moffitt and Bruce Tate) for programmers new to Julia, which is very good, with nice examples and exercises.
 
+# Upcoming
+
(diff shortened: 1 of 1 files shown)


ViralBShah:
What does early access mean? Is it possible to buy? If so, we should merge.

ViralBShah:
Perhaps say a bit about the Early Access bit? I am a little concerned that the first book link points to a book that will be out next summer.

How about make a separate section right below, called upcoming books, and link it there. The link could also point to the explanation of MEAP.

Ismael-VC:
:+1: for upcomming books, there is also Lea's one.

chrisvoncsefalvay:
> What does early access mean? Is it possible to buy? If so, we should merge.

It's possible to buy, yes. It means you get 3 chapters now and a chapter every month or so. The final print version will be available next Summer, but if you buy now, you will get the book as it comes out, chapter by chapter.

chrisvoncsefalvay:
> upcoming books

Not necessarily opposed to it, but in my slightly pedantic understanding, a book is 'upcoming' if it has not been published yet. It is no longer 'upcoming' once it's in Early Access or some other form of public beta. Gmail, which was in public beta for half a decade, was not 'upcoming software'. :)

But that's really just my pedantry.

ViralBShah:
I think it is upcoming so long as the full book is not available! The book description can clearly include what is available already.

jiahao:
I agree with @ViralBShah; some explanation of Early Access would be helpful.

johnmyleswhite:
Let's get something like this change merged soon. We should explain Early Access very briefly, but it's reasonable to assume anyone can figure out what the term means given that essentially all modern tech book publishers provide such programs.

jiahao:
I would be fine with simply adding a link describing Manning's Early Access policy. to this entry. Perhaps @Aleks-d can provide a link to such a page?

Aleks-d:
Hi. You can find more details about Manning Early Access Program [here](http://www.manning.com/about/meap.html).

jiahao:
@chrisvoncsefalvay please add the link and squash the commits and then I think it should be ready.

chrisvoncsefalvay:
Should be mergeable now, @jiahao ;) Thanks @Aleks-d  for the Meap link.

PR #258, opened 2015-08-21
added a few more plotting packages

cormullion:
Ref: https://github.com/dcjones/Gadfly.jl/issues/658

### DIFF
--- a/downloads/plotting.md
+++ b/downloads/plotting.md
@@ -5,15 +5,17 @@ title:  Plotting in Julia
 
 # Plotting in Julia
 
-Plotting in Julia is available through external packages.
+Plotting in Julia is available through external packages. This is a relatively complete list of packages available on GitHub that provide plotting capabilities as Julia packages. Most are still under active development.
 
-## PyPlot
+## Graphical plotting packages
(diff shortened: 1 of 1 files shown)


ihnorton:
This page is not intended to be a grab bag of plotting libraries, but rather a recommendation of a small number of the best options for new users. The proposed list does not seem to have been even minimally vetted, as it lists things like julia-plot which has not been maintained in *years*, and several others that are clearly not appropriate to recommend to new users.

ihnorton:
A few thoughts on the proposed list:

Yes: Winston
Maybe: one of Bokeh or Vega

No (non-local, requires external accounts): Plotly, Google Charts
No (insufficient benefit over other options): Gaston, QWT
No (not yet generally usable): GLPlot
No (too niche / minimal / unmaintained): TextPlots, ASCIIPlots, Sparklines, Hinton, ImageTerm, julia-plot

ViralBShah:
I would not mind having a different page that lists all of these, or a blog post that compares all these, that we link - but this page should just recommend 2-3 different and well functioning options.

PR #260, opened 2015-09-07
Create mixin for videos from JuliaCon (closes #257)

### LINKED ISSUE #257: Load a random JuliaCon video on front page
I think it would be nice to have a random JuliaCon 2015 video load on the front page of julialang.org. The content and recordings are both high quality.

What do folks think? Also, it would be great if someone knows what javascript to put in to make this magic happen.

tlycken:
This creates a mixin that can be used to include a random video from the JuliaCon 2015 playlist on Youtube. It's usage is exemplified by the changes to `index.md`.

### DIFF
--- a/index.md
+++ b/index.md
@@ -11,6 +11,8 @@ In addition, the Julia developer community is contributing a number of [external
 Julia programs are organized around [multiple dispatch](http://docs.julialang.org/en/release-0.3/manual/methods/#man-methods); by defining functions and overloading them for different combinations of argument types, which can also be user-defined.
 For a more in-depth discussion of the rationale and advantages of Julia over other systems, see the following highlights or read the [introduction](http://docs.julialang.org/en/release-0.3/manual/introduction/) in the [online manual](http://docs.julialang.org).
 
+{% include juliacon-player.md %}
+
 # A Summary of Features
 
 * [Multiple dispatch](http://en.wikipedia.org/wiki/Multiple_dispatch): providing ability to define function behavior across many combinations of argument types
--- a/index.md
+++ b/index.md
@@ -11,8 +11,6 @@ In addition, the Julia developer community is contributing a number of [external
 Julia programs are organized around [multiple dispatch](http://docs.julialang.org/en/release-0.3/manual/methods/#man-methods); by defining functions and overloading them for different combinations of argument types, which can also be user-defined.
 For a more in-depth discussion of the rationale and advantages of Julia over other systems, see the following highlights or read the [introduction](http://docs.julialang.org/en/release-0.3/manual/introduction/) in the [online manual](http://docs.julialang.org).
 
-{% include juliacon-player.md %}
-
 # A Summary of Features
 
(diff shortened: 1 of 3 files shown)


ViralBShah:
This is great. Would like a couple others to review this too and merge.
Cc:@shashi

index.md
```diff
@@ -11,6 +11,8 @@ In addition, the Julia developer community is contributing a number of [external
 Julia programs are organized around [multiple dispatch](http://docs.julialang.org/en/release-0.3/manual/methods/#man-methods); by defining functions and overloading them for different combinations of argument types, which can also be user-defined.
 For a more in-depth discussion of the rationale and advantages of Julia over other systems, see the following highlights or read the [introduction](http://docs.julialang.org/en/release-0.3/manual/introduction/) in the [online manual](http://docs.julialang.org).
 
+{% include juliacon-player.md %}
```
shashi: Can probably add a title / hint as to what is being shown here...

shashi:
looks good to me!

index.md
tlycken: Sure - I thought about that, but figured someone else can probably make a better decision than me on what to write (and where to place this).

As a matter of fact, I think I'll add a commit that removes this addition, and then this PR can be merged without changing the look of the site. Then, someone else can write a full section of text and use this mixin to do the actual inclusion of the video.

tlycken:
I now removed the actual usage of the mixin, so this can be merged without changing how anything looks on the site.

To add an embedded youtube player that picks a random video from the JuliaCon 2015 playlist, just add 

    {% include juliacon-player.md %}

on its own line in one of the markup files.

ViralBShah:
The image that is shown is always of the first video. Is it possible to have it show the image of the randomly presented video?

tlycken:
I don't know - that image is part of what Youtube's embedded player gives us, so I doubt it can be done easily. I could attempt something based on http://stackoverflow.com/a/16751464/38055; in other words, instead of creating an `<iframe>` with content directly from Youtube, create an `<object>` with media linked from Youtube (which is basically what the `<iframe>` will contain). I have some other stuff I need to attend to now, though, but I can take a stab at it tomorrow.

It should be possible to change just the mixin, though, so if you want to start using this version I can create a PR with the next one tomorrow and it will just update itself :)

ViralBShah:
I am already using this version. Thanks!

tlycken:
I've looked at the documentation and searched a little, but I can't find a way to fix the thumbnail issue. We'll see if http://stackoverflow.com/questions/32451701 yields anything - if not, we'll just have to live with it I guess.

shashi:
It looks like the only way to do this is to enumerate all the videos.

I did something that works, see: #261. The down side is we might need a new mixin for every playlist.

tlycken:
> we might need a new mixin for every playlist.

Not necessarily - [mixins can take parameters](https://jekyllrb.com/docs/templates/#includes) so we might be able to create a single mixin that picks a random video from a set of ids, and then pass the ids as a parameter where used.

But it still won't be as pretty as it could if Youtube would just give us the right thumb to start with. I've filed a bug report (or at least I think I have, via their "feedback" feature) about this.

shashi:
Sounds good!

PR #261, opened 2015-09-08
Show the correct cover picture in the JuliaCon random video

shashi:


ViralBShah:
Nice. This works, even though will have to be redone for every playlist.

PR #262, opened 2015-09-09
[WIP] Prepare v0.4.0-rc1 download linkage

staticfloat:
We've done a pretty nice job of making the download page sell some of our strong integrations like Juno, JuliaBox, etc... do we want to make a Juno bundle for 0.4-rc1?  Do we want to ensure that JuliaBox is up and running with 0.4-rc1 before announcing further?

Pinging @one-more-minute @ViralBShah directly for these points.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -95,6 +95,41 @@ the Juno executable or app.
 
 ---
 
+# Release Candidates (v0.4.0-rc1)
+
+We are currently testing release candidates for the next major version of Julia.  Please see
+[platform](platform.html) specific instructions if you have trouble installing Julia.
+Checksums for this release are available in both
+[MD5](https://s3.amazonaws.com/julialang/bin/checksums/julia-0.4.0-rc1.md5)
+and [SHA256](https://s3.amazonaws.com/julialang/bin/checksums/julia-0.4.0-rc1.sha256) format.
+Download the release candidate here, and please report any issues to the
+[issue tracker](https://github.com/JuliaLang/julia/issues) on Github.
+
+## Julia (command line version)
+<table class="downloads"><tbody>
+<tr>
+    <th> Windows Self-Extracting Archive (.exe) </th>
+    <td> <a href="https://s3.amazonaws.com/julialang/bin/winnt/x86/0.4/julia-0.4.0-rc1-win32.exe">32-bit</a> </td>
+    <td> <a href="https://s3.amazonaws.com/julialang/bin/winnt/x64/0.4/julia-0.4.0-rc1-win64.exe">64-bit</a> </td>
+</tr>
+<tr>
+    <th> Mac OS X Package (.dmg) </th>
+    <td colspan="2"> <a href="https://s3.amazonaws.com/julialang/bin/osx/x64/0.4/julia-0.4.0-rc1-osx10.7+.dmg">10.7+ 64-bit</a> </td>
+</tr>
+<tr>
+    <th> Generic Linux binaries </th>
+    <td> <a href="https://julialang.s3.amazonaws.com/bin/linux/x86/0.4/julia-0.4-rc1-linux-i686.tar.gz">32-bit</a> </td>
+    <td> <a href="https://julialang.s3.amazonaws.com/bin/linux/x64/0.4/julia-0.4-rc1-linux-x86_64.tar.gz">64-bit</a> </td>
+</tr>
+<tr>
+    <th> Source </th>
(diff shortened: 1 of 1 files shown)


downloads/index.md
```diff
@@ -95,6 +95,41 @@ the Juno executable or app.
 
 ---
 
+# Release Candidates (v0.4.0-rc1)
+
+We are currently testing release candidates for the next major version of Julia.  Please see
```
tkelman: minor

staticfloat: I'm looking for the reference, but I remember @StefanKarpinski saying somewhere that before 1.0, versions such as `0.4` and `0.3` were considered "major" versions.  Since there does seem to be some confusion regarding this, I'll just remove "major" and say "the next version of Julia".

tkelman:
Are these good to go? Should we send an announcement to julia-users and julia-news? I can follow it up with specific Travis instructions if you want to write the main email.

staticfloat:
Sure, but I'd like to see if Viral or Mike think it's worthwhile packaging Juno or doing some JuliaBox stuff before then.  Unified front and all that.  Let's let it sit for a few hours.

ViralBShah:
I think we should not hold the wider RC1 announcement for Juno and JuliaBox to catch up. They should certainly do so by RC2.

I wonder if Juno can build automatically like the rest of Julia. That would make things easier. Hopefully, we will have moved to Juno (Atom) by 0.5, so this may be the last Juno (LT) release - but we'll have to support it for every minor release on 0.4.

ViralBShah:
Just occurred to me to ask whether there is a way to sign a linux tarball. Would be nice to do that.

tkelman:
Could use gpg ?

staticfloat:
That we could, and publish the signatures in the same way as the checksums.  Signing it would make the checksums redundant, but I don't mind making it easy for people to trust us.  :)

nalimilan:
I think we should give the link to the tarball resulting from `make light-source-dist` instead of the GitHub tarballs (on the webpage as well as on the GitHub release page). The former doesn't require installing Sphynx to build the documentation, which is much simpler for packagers (as well as for all users in general).

staticfloat:
We give the link to the tarball resulting from `make full-source-dist`.  Doesn't that work?

jiahao:
Ok we have a green light from both Mike and Viral. Merge away!

one-more-minute:
Yes, I'll be able to update Juno over the next couple of days and check it's all working properly with 0.4. Until then people can point Juno at 0.4 fairly easily if they want to try out the beta release, so I agree that it shouldn't block things.

ViralBShah:
I like the term - united front!

nalimilan:
@staticfloat No, because distro packagers don't want all the dependencies included in Julia source packages. That would make them ten times bigger as needed. All projects that I know of publish a tarball with only their sources, bundling dependencies is only a bonus.

staticfloat:
Ah, I see what you mean.  Yeah, we could host one of those somewhere I suppose.  :P

nalimilan:
Can I help? Since you need to sign the tarball, I'm not sure there's much I can do, but feel free to tell me the contrary.

staticfloat:
Thanks for the offer, I've already done it.  :)

nalimilan:
Thanks, I'm going to update the Fedora package based on it.

FWIW, in the future, it would be nice to remove the commit version from the tarball name. That way, packagers only need to change the version in their .spec file (or similar), and the tarball will be downloaded directly from the updated URL. Fedora is even connected with https://release-monitoring.org now, and packages can be built automatically when all is needed is changing the version number. This means minor versions (at least) can be packaged within minutes after their release.

ViralBShah:
That is a good idea. release-monitoring.org is a great idea - didn't know about that.

staticfloat:
@tkelman I believe we had some reason for why we wanted the `gitsha` in the tarball filename originally; something about wanting to know the Julia version inside quickly and easily.  Nightlies should almost certainly continue to have gitshas in them, but now that the Julia releases have their gitshas baked into them (via `build_h.jl` and friends) do you see any reason not to comply with what https://release-monitoring.org wants?

nalimilan:
Indeed nightlies should keep the commit in the filename, that's their only identifier. My proposal only applies to releases.

(BTW, this is not only for https://release-monitoring.org, which is quite new and Fedora-specific at this point: all packagers would be happy to be able to update the URL programmatically, without even looking at the tarball name.)

tkelman:
I think it's because the `binary-dist` targets do double-duty for both nightlies and releases. I don't see an issue with manually renaming the files before we upload them for releases.

PR #263, opened 2015-09-10
Add GPG signing of Linux tarballs and source tarball

staticfloat:
Yay cryptography.


ViralBShah:
Awesome!

nalimilan:
The table looks weird here, with parentheses on a different lines than `GPG signature`. Is that intended?

tkelman:
Oh, that is ugly. It looks okay in markdown?

PR #280, opened 2015-10-03
JSoC '15 NullableArrays blog post

davidagold:
Here is my JSoC blog post. Do let me know if there are any changes I ought to make!

Thank you,
David


johnmyleswhite:
Looking pretty good to me.

Not sure I understand the labels on the plots. I see "none", "~half" and "N/A". Is "N/A" equivalent to "all"?

davidagold:
Technically "N/A" is "none" in the sense that regular Arrays cannot hold
null values so it doesn't make sense to ask after the proportion of null
values in a regular Array. But really, the only reason I included such a
level is because if "none" is used then the data point for regular Arrays
hides the data point for NullableArrays in many cases, and I wasn't able to
jitter the data points in the x direction. If you've a suggestion for
overcoming that representational hurdle I'd be happy to use that rather
than the three level solution, which I agree is kind of confusing.

ScottPJones:
N/A normally means not not applicable, and isn't really the same as Null
It's a fine distinction which frequently isn't made correctly.
Null means "there is no answer", NA means "we know the answer, and it's not applicable".

StefanKarpinski:
To quote [the R manual](https://cran.r-project.org/doc/manuals/r-release/R-lang.pdf):

> NA is used for absent (“Not Available”) data values.

Classic SPJ – so close but yet so far.

ScottPJones:
Look closer, and stop with the disrespectful comments.

ScottPJones:
Here's a good explaination: https://pds.jpl.nasa.gov/documents/sr/Chapter17.pdf

johnmyleswhite:
@ScottPJones: Please stop commenting. You are going to make things much worse for yourself if you keep talking.

ViralBShah:
I'd like to delete all the comments here after my publication of the blog post. This discussion as many others is not useful, and this is not the place for it.

PR #294, opened 2015-10-26
add USTC Julia User Group to the local user groups

Roger-luo:
this local user group use QQ group and ID number is 316628299
maybe can be written like 

USTC Julia User Group,QQ Group:316628299

### DIFF
--- a/blog/_posts/2015-10-25-datastreams.md
+++ b/blog/_posts/2015-10-25-datastreams.md
@@ -1,6 +1,6 @@
 ---
 layout: post
-title:  JSoC project update - DataStreams.jl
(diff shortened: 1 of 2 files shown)


ViralBShah:
This PR doesn't seem to include that commit. Perhaps needs redoing?

Also, that text is very cryptic. Can you please provide some elaboration and URLs for the user group?

Roger-luo:
sorry,I'm not sure which branch should I put this PR. and 
USTC Julia User Group is a user group in the university of Science and Technology of China,
it's not convenient to access google groups in China so we use a QQ group for discussing at present.

PR #295, opened 2015-10-26
add qq groups to community

Roger-luo:
qq gourps is a kind of thing like chat room in China,more convenient if there is one in China

add QQ groups to the community
add USTC(university of Science and Technology of China) Julia User Group,QQ group:316628299 in QQ groups


StefanKarpinski:
This is a little hard to understand – could you maybe provide a little more context or links?

Roger-luo:
@StefanKarpinski umm, which part is hard to understand? 

The QQ group is something like chat room in China.QQ is a chat service of Tencent company. 

And it's hard to approach Google services in China .So mail list could be hard to use for inner Chinese users.

We have not found any other services which may have similar functions to Google groups. And it's more common to use QQ groups in China :-)

The QQ groups do not offer any link that allows people to join in directly. so a group ID is needed(which may make the page looks ugly?)

And we are developing a website for Chinese community so there would be a link in the future:-)

Any advice on how to add our community connections to the page?

StefanKarpinski:
A link to QQ would help http://qqchat.qq.com/ with some instructions for how to use the number maybe?

PR #297, opened 2015-10-27
update qq groups in community

Roger-luo:
update descriptions about qq services and add some other qq groups to the community hope this would help :-)


ViralBShah:
Thanks, this is much better to understand for people who are not familiar with Tencent and QQ.

PR #301, opened 2015-11-09
Update download links for 0.3.12 and 0.4.1

tkelman:
This is missing the gpg signatures which only @staticfloat can currently create with the existing key.


tkelman:
gpg details communicated offline. looks like it's all working.

staticfloat:
Awesome, nice job @tkelman!

PR #303, opened 2015-11-25
bump juno version

one-more-minute:
This bumps Juno to the latest Julia bugfix release. Unfortunately, these builds aren't signed yet – I'm not sure whether an unsigned newer release is better than a signed older one. We should be able to fix that within the next couple days, though.


ViralBShah:
I think we can update it once signed.

Issue #307, opened 2015-12-07
Gemfile missing


dancinglightning:
Hi,
i just cloned to see if the docs.julilang.org are also in this repository.
Following README (i also use jekyll on github sites) i 
> bundle install

to find there is no Gemfile. This is really short 

source 'https://rubygems.org'
gem 'github-pages'

The created Gemfile.lock should be checked to avoid people having different versions.
I can off course create a pull but i guess it's such a small thing. 

About my original question: where are the docs.julialang.org ? I have a small suggestion there.

Torsten

pao:
Please feel free to make a pull request to add the Gemfile--it's a small thing, but it's a small thing you know how to do!

The documentation is generated from files in the main Julia repository.

one-more-minute:
Looks like we can close this with #308 merged.

PR #308, opened 2015-12-08
add Gemfile and small reformat of readme

dancinglightning:
just made the readme a little more readable, no real change, just formatting
Gemfile as per issue #307


pao:
LGTM, thanks!

ViralBShah:
Do we need the Gemfile.lock?

dancinglightning:
While the Gemfile says which gems u uses, the lock file says which versions are actually used.
Checking it in means everyone uses the same versions and the maintainer does not have to chase weird bugs that come from using different versions.
i'd go with yes :-)

ViralBShah:
Thank you for the explanation. I was just wondering if it was a temporary lockfile that got checked in by mistake.

PR #309, opened 2015-12-08
small layout tweaks

dancinglightning:
tiny bit wider main layout, stops the menu breaking
wider blogindex stops the post names breaking,
removed nil layout to stop jekyll complaining


pao:
cc @ViralBShah regarding the style changes

PR #311, opened 2015-12-15
julia.bat is no more

PallHaraldsson:
I'm assuming julia.exe ..

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -11,7 +11,7 @@ Julia is available for both 32-bit and 64-bit Windows since XP SP2.
 
 1. Download the Windows julia.exe installer for your platform. 32-bit julia works on both x86 and x86_64. 64-bit julia will only run on 64-bit Windows (x86_64).
 2. Run the downloaded program to extract julia
-3. Double-click julia.exe in the unpacked folder to start julia
+3. Double-click the julia shortcut in the unpacked folder (or run it's target julia.exe in the bin/ sub-folder) to start julia
 
 The [Windows README](https://github.com/JuliaLang/julia/blob/master/README.windows.md) contains information on dependencies.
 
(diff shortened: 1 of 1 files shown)


tkelman:
Not quite, there's no julia.exe in the top-level install folder (it's under bin/julia.exe), but there is a julia shortcut.

PallHaraldsson:
I hope the test is ok and ok not to squashing the commits.. I think I'll figure out rebase and squash (I need  to for elsewhere, it seems I need git from the shell, with github.com not having some user interface..).

tkelman:
remove the parenthetical, it's not really necessary

tkelman:
I meant remove the note that was inside the parenthetical, not just the () around it. Most users don't need to worry about the fact that the shortcut points to bin/julia.exe.

PallHaraldsson:
@tkelman I could do that.

I think the shortcut is only useful to start Julia/the REPL. I actually will be experimenting with running Julia scripts on Windows (machine I have to borrow..). In Linux #!/bin/julia takes care of running scripts. I noticed to run a script in Windows the extension .jl doesn't trigger running the program. Maybe I did something wrong (I know I could probably register the extension, but should I need to).

Anyway, I assume in Windows you would do:

[maybe longer path]bin/julia.exe my_script.jl

would it not help just to have it in? I think you can not do julia.shortcut [or whatever] my_script.jl.

Yes, most users would know this or?

tkelman:
Windows doesn't do shebangs for script execution. So you can create your own shortcuts if you'd like.

The grammar is also wrong, should be its not it's.

tkelman:
Closed by f2e4dc5471f106404c1e69b6cbd0412f85be0a34. These are basic "getting started" instructions, we can discuss creating your own shortcut and making use of `bin/julia.exe` somewhere else.

PR #312, opened 2015-12-16
Don't claim benchmarks are idiomatic

johnmyleswhite:
The assertion that we use a "reasonable idiom in each language" is too easily argued against. I believe that phrase doesn't help sell the benchmarks. If you believe that, I think this change helps to provide one less avenue for attack.


StefanKarpinski:
Good call.

ViralBShah:
Yes, this is definitely the right thing.

PR #313, opened 2015-12-22
Add reference to SINTEL 2015 workshop.

OscarNeciosup:


ViralBShah:
Merging for now, but perhaps for conferences, we should have a separate section?

OscarNeciosup:
I think that's good , first time there is an event in my country about
Julia . Thanks for passing this.

2015-12-22 5:12 GMT-05:00 Viral B. Shah <notifications@github.com>:

PR #322, opened 2016-01-29
Fix logo dot alignment

simonster:
- Blue, red, and purple dots now have same y position
- Make red and purple dots have the same distance from the center of the green dot


jiahao:
This change may require Viking inspection.

simonster:
bump

PR #324, opened 2016-01-31
Add blog post on multidimensional iteration and indexing

timholy:
Summarizes the work of @mbauman, @Jutho, and myself on indexing and iteration, in an attempt to make this information more widespread.

Is there an easy way to preview it? I checked it using a locally-installed markdown renderer, and it looked fine, but that doesn't use all the nice highlighting.

### DIFF
--- a/blog/_posts/2016-02-01-iteration.md
+++ b/blog/_posts/2016-02-01-iteration.md
@@ -1,0 +1,420 @@
+---
+layout: post
+title:  Multidimensional algorithms and iteration
+author: <a href="http://holylab.wustl.edu">Tim Holy</a>
+---
+
+Starting with release 0.4, Julia makes it easy to write elegant and
+efficient multidimensional algorithms. The new capabilities rest on
+two foundations: a new type of iterator, called `CartesianIndex`, and
+sophisticated array indexing mechanisms.  Before I explain, let me
+emphasize that developing these capabilities was a collaborative
+effort, with the bulk of the work done by Matt Bauman (@mbauman),
+Jutho Haegeman (@Jutho), and myself.
+
+These new iterators are deceptively simple, so much so that I've never
+been entirely convinced that this blog post is necessary: once you
+learn a few principles, there's almost nothing to it.  However, like
+many simple concepts, the implications can take a while to sink in.
+There also seems to be some widespread confusion about the
+relationship between these iterators and
+[`Base.Cartesian`](http://docs.julialang.org/en/stable/devdocs/cartesian/),
+which is a completely different (and much more painful) approach to
+solving the same problem.  There are still a few occasions where
+`Base.Cartesian` is necessary, but for many problems these new
+capabilities represent a vastly simplified approach.
+
+Let's introduce these iterators with an extension of an example taken
+from the
+[manual](http://docs.julialang.org/en/stable/manual/arrays/#iteration).
+
+## `eachindex`, `CartesianIndex`, and `CartesianRange`
+
+You may already know that, in julia 0.4, there are two recommended
+ways to iterate over the elements in an `AbstractArray`: if you don't
+need an index associated with each element, then you can use
+
+```jl
+for a in A    # A is an AbstractArray
+    # Code that does something with the element a
+end
+```
+
+If instead you also need the index, then use
+
+```jl
+for i in eachindex(A)
+    # Code that does something with i and/or A[i]
+end
+```
+
+In some cases, the first line of this loop expands to `for i =
+1:length(A)`, and `i` is just an integer.  However, in other cases,
+this will expand to the equivalent of
+
+```jl
+for i in CartesianRange(size(A))
+    # i is now a CartesianIndex
+    # Code that does something with i and/or A[i]
+end
+```
+
+Let's see what these objects are:
+
+```jl
+A = rand(3,2)
+
+julia> for i in CartesianRange(size(A))
+           @show i
+       end
+i = CartesianIndex{2}((1,1))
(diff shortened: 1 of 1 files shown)


jiahao:
:+1: Reads well to me

tkelman:
Nice! I definitely learned something here.

jakebolewski:
Great Post!  I think you have to give a brief definition of type-instability (as you mention it often) or link to other material that explains this concept more in depth.  This may be a common term thrown around in the Julia community, but there is no clear definition for this concept and it is not widely used outside it.

blog/_posts/2016-02-01-iteration.md
```diff
@@ -0,0 +1,420 @@
+---
+layout: post
+title:  Multidimensional algorithms and iteration
+author: <a href="http://holylab.wustl.edu">Tim Holy</a>
+---
+
+Starting with release 0.4, Julia makes it easy to write elegant and
+efficient multidimensional algorithms. The new capabilities rest on
+two foundations: a new type of iterator, called `CartesianIndex`, and
```
mbauman: Nitpick: the iterator is `CartesianRange`, which has `CartesianIndex` elements.

blog/_posts/2016-02-01-iteration.md
```diff
@@ -0,0 +1,420 @@
...
+CartesianRange{CartesianIndex{2}}(CartesianIndex{2}((-7,0)),CartesianIndex{2}((7,15)))
+```
+
+constructs a range that will loop over `-7:7` along the first
+dimension and `0:15` along the second.
+
+One reason that `eachindex` is recommended over `for i = 1:length(A)`
+is that some `AbstractArray`s cannot be indexed efficiently with a
+linear index; in contrast, a much wider class of objects can be
+efficiently indexed with a multidimensional iterator.  (SubArrays are,
```
mbauman: Do you think it would be worth talking about the asymmetry here?  We wouldn't have done this if x86 div/rem were as fast as multiplication and addition… and it's something that definitely isn't obvious.  I'm not sure you need to get too deep into the weeds, but it might be worth simply saying that computers can very efficiently use a multidimensional index for all kinds of arrays, but going the other way can be really slow.

blog/_posts/2016-02-01-iteration.md
```diff
@@ -0,0 +1,420 @@
...
+  output (what if the input array element type `T` is `Int`?), but
+  we're cutting a few corners here for simplicity.
+
+- `I1 = CartesianIndex(ntuple(d->1, N)::NTuple{N,Int})` creates an
+  `N`-dimensional `CartesianIndex` filled with 1s, e.g.,
+  `CartesianIndex{3}(1,1,1)` in three dimensions.  The type
+  annotation, `::NTuple{N,Int}` is unfortunately necessary (if you
+  want good performance), because julia 0.4 cannot infer the output
+  type of (most) calls to `ntuple`.  We'll explore this topic further
+  below.
```
mbauman: One way around introducing type annotations this early on would be:

```jl
R = CartesianRange(size(A))
I1 = first(R)
Iend = last(R)
for I in R
```

I think it's more readable, too.

mbauman:
:+1:.

blog/_posts/2016-02-01-iteration.md
tkelman: The `first` / `last` version looks nicer to me too.

Jutho:
Great read. `CartesianRange/Index` have already evolved significantly since the first version to which I contributed. Thanks for all the great work @timholy and @mbauman !

blog/_posts/2016-02-01-iteration.md
```diff
@@ -0,0 +1,420 @@
...
+author: <a href="http://holylab.wustl.edu">Tim Holy</a>
+---
+
+Starting with release 0.4, Julia makes it easy to write elegant and
+efficient multidimensional algorithms. The new capabilities rest on
+two foundations: a new type of iterator, called `CartesianIndex`, and
+sophisticated array indexing mechanisms.  Before I explain, let me
+emphasize that developing these capabilities was a collaborative
+effort, with the bulk of the work done by Matt Bauman (@mbauman),
+Jutho Haegeman (@Jutho), and myself.
```
mauro3: Add `@timholy`?

rfourquet:
Thank you!

blog/_posts/2016-02-01-iteration.md
nalimilan: Also, by "we'll explore this topic", I thought you were going to speak of `ntuple`'s type instability. You could say "we'll explore the issue of type instability" instead.

timholy: Great idea about using `first` and `last` here. Being able to use `first` to compute the offset range is, of course, a trick that's very specific to the fact that we're doing 3-by-3-by... (it wouldn't work if it were 5-by-5-by...). But it **is** the best way to handle the enforcement of the bounds, so in that respect this is an improvement in terms of generality.

As you say, it would also be better to avoid introducing type instability so early, so for pedagogical reasons I'll take all the dirty tricks I can get! Thanks!

timholy:
Thanks for the comments everyone, they were very helpful. I'll leave this open a bit longer in case others want to chime in.

blog/_posts/2016-02-01-iteration.md
timholy: GitHub didn't auto-close this from the commit, but I addressed this too, by linking to [devdocs](http://docs.julialang.org/en/stable/devdocs/subarrays/#indexing-cartesian-vs-linear-indexing).

PR #327, opened 2016-02-06
Highlight code in `iteration` blog post.

Carreau:
I enjoyed @timholy post , 

I prefer it highlighted.  (only drawback is the style of some unicode that are red-ish
screen shot 2016-02-06 at 15 00 23

You should be able to [view the diff without whitespace changes](https://github.com/JuliaLang/julialang.github.com/pull/327/files?w=1) by appending `?w=1`

### DIFF
--- a/blog/_posts/2016-02-01-iteration.md
+++ b/blog/_posts/2016-02-01-iteration.md
@@ -34,38 +34,46 @@ You may already know that, in julia 0.4, there are two recommended
 ways to iterate over the elements in an `AbstractArray`: if you don't
 need an index associated with each element, then you can use
 
-     for a in A    # A is an AbstractArray
-         # Code that does something with the element a
-     end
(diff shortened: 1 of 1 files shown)


tkelman:
Makes sense to me?

StefanKarpinski:
Thanks, @Carreau!

timholy:
Just noticed this. Thanks for fixing it!

timholy:
I'm still not seeing highlighting when I visit the page, though. Any ideas?

Carreau:
> I'm still not seeing highlighting when I visit the page, though. Any ideas?

Hum, no, if I just run jekyll locally I get the highlighting.

Might be related to recent github update:

https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

PR #329, opened 2016-02-10
Update Summer of Code info

shashi:
In preparation for a GSoC 2016 ideas page, I have merged the 2015 and 2014 ideas page into a new central ideas page that we can keep updated all year round.

I have set up redirects for the old pages to this one.


shashi:
Fixed merge conflicts.

PR #330, opened 2016-02-10
better typography and navbar

shashi:
Used Crimson Text font and a size scale for headers.

The navbar now shows where you are guiding you around better. Harmless easing animation on hover.


StefanKarpinski:
Definitely an improvement.

PR #332, opened 2016-02-11
Add constrained optimization, tweak CUTEst description

timholy:
CC @mlubin (edit at will).


mlubin:
CC @dpo, are you developing a set of benchmark instances in JuMP?

dpo:
The official CUTEst interface is now at https://github.com/JuliaOptimizers/CUTEst.jl. The one you're pointing to is deprecated. It should probably be removed, unless there's a compelling reason to keep it around @abelsiqueira.

I currently have a few unconstrained problems in JuMP: https://github.com/JuliaOptimizers/OptimizationProblems.jl. The idea is to add to that collection, but time's been short lately.

abelsiqueira:
No reasons, it is deprecated.

timholy:
If you're on Linux, where do you get `sifdecoder`?

abelsiqueira:
[homebrew-cutest](https://github.com/optimizers/homebrew-cutest) is currently the best option, even on Linux using [Linuxbrew](https://github.com/Linuxbrew/linuxbrew), because it compiles correctly the shared libraries.

timholy:
Didn't know about Linuxbrew, that's very helpful.

I'm presuming that the remaining road bumps in test suites will not need a whole standalone project, so I merged the "constrained optimization" and "standard test suite" into a single project. Any more comments, or should I merge?

mlubin:
LGTM if we can find the right student.

PR #333, opened 2016-02-13
added sciences po paris dept of economics 

floswald:
computational economics course with julia.


jiahao:
Thanks, but please follow the existing style. Institution on first level bullet point, course on second level.

floswald:
ok, on two lines now.

Issue #319, opened 2016-01-25
make error in publications directory 


weijianzhang:
Hi, 

When I ran `make` in the `pulications` directory, I got the following error:
```
$ make
PATH=/home/weijian/anaconda3/bin:/home/weijian/anaconda/bin:/home/weijian/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/bin pandoc --reference-links -s -S --biblio julia.bib --csl jlwebpage.csl -o index_new.html _EDIT_ME_index.md
sed: can't read : No such file or directory
make: *** [all] Error 2
```
It seems `sed` can not find `index_new.html`. Note I have both pandoc and pandoc-citeproc installed.

Could you let me know how I can fix this error? 
 
Thanks a lot,

Weijian

jiahao:
It looks like the sed syntax I used only works on BSD/Darwin sed, not on Linux. I'll commit the fix after I check that it still works on OSX.

weijianzhang:
Thanks a lot.

jiahao:
Fixed by #338

PR #338, opened 2016-02-16
Publications makefile: Linux-compatible sed syntax

jiahao:
Apparently BSD sed allows a space after -i but Linux sed does not.

TODO: verify that this change does not break on Mac OS X

Ref: #319


tkelman:
it probably does break osx, i think you can write an explicit extension like `-ibak`

PR #342, opened 2016-02-18
convex.jl project idea

mlubin:
Copied from https://github.com/numfocus/gsoc/blob/master/2016/ideas-list-juliaopt.md
CC @madeleineudell


shashi:
Cool! :)

PR #346, opened 2016-02-22
Helper functions for CUDA Programming

ChrisRackauckas:
A proposal for CUDA helper functions. Includes step-by-step things that could be done. I would offer to help guide the project if needed.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -182,6 +182,16 @@ Some possible aims of this project:
 * Efficient generation of non-uniform variates, across different floating point precisions.
 * Massively parallel random number generators, such as [SPRNG](http://www.sprng.org) or the [Random123](https://www.deshawresearch.com/resources_random123.html) entropy streams, and integration with [ComputeFramework.jl](https://github.com/shashi/ComputeFramework.jl).
 
+## Helper functions for CUDA Programming
+
+CUDA has become a major part of high-performance computing. Julia currently supports CUDA programming via the CUDA.jl package, but its usability since the interface is very low level. The goal of the project would be to implement wrappers for basic CUDA programs in order to help facilitate CUDA programming in Julia. Possible aims for the project include:
+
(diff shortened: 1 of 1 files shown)


ViralBShah:
Between Cudart.jl and https://github.com/JuliaComputing/ArrayFire.jl, don't we have much of this? Perhaps this project proposal should build on these further.

ChrisRackauckas:
Is there a way to have persistent memory for CUDArt.jl? I don't see an example in the tests, so I assumed it wasn't possible.

I was thinking the intuitive way would be to have persistent arrays on the GPU that just kind of act like "normal" arrays, but in the backend have wrappers that run on the GPU. If the memory isn't persistent, then the overhead of transferring is too much for HPC use. If this is possible via CUDArt.jl, then yes extending that package would be the way to go.

timholy:
First, CUDA.jl is effectively deprecated, you probably want CUDArt.jl. https://github.com/JuliaGPU/CUDA.jl/pull/16#issuecomment-185558852

Yes, you can have persistent arrays. You just have to handle the resource initialization and cleanup manually: `CUDArt.init(devlist)` followed by your code, then `CUDArt.close(devlist)` when you are done.

Agreed that we want to compile julia functions to CUDA kernels, but IIUC there's quite a ways to go to get that working. Or are you planning on a different route than via Julia->LLVM->PTX?

ChrisRackauckas:
Thanks for notifying me about CUDA.jl. I just spent some time switching a lot of my current code over to CUDArt.jl. Using init like you said and launching the functions directly worked well. I will write a blog tutorial on this once my blog is back up.

And I was thinking of Julia->LLVM->PTX. LLVM does have a way of compiling to PTX, so I thought that getting that going in Julia would be a main priority, and once that goes through the rest is just using standard Julia functions through this path to make all the default behavior. But if you think that's too tough to handle now, I'll trust your judgement.

ViralBShah:
Cc @timholy @maleadt @SimonDanisch

ViralBShah:
There is an open PR for julia->llvm->ptx.

PR #347, opened 2016-02-24
Updated GSOC ODE.jl

mauro3:
Cc: @ViralBShah


ViralBShah:
It is preferable not to explicitly model on the matlab API, and instead go with something newer that the community wants, or is Julian. Could you update that?

mauro3:
You're right.  Worded it more generically, better?

PR #348, opened 2016-02-27
Add mention of the Julia Computing blog on the Blog page

nalimilan:
There have been very interesting posts there recently.

### DIFF
--- a/blog/index.html
+++ b/blog/index.html
@@ -3,7 +3,7 @@ layout: default
 title:  The Julia Blog
 ---
 <div id="blogindex">
-  <p>The Julia blog discusses issues of numerical, technical, distributed and parallel computing, as well as programming language design, and how these issues touch upon the design and implementation of the Julia programming language.</p>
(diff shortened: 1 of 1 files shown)


ViralBShah:
The Julia Computing Blog was registered with juliabloggers.com just a couple of days ago. We have a few more coming, and hope to keep it up - and will also be publishing more blogs on the julialang blog.

nalimilan:
Yes, I've seen that, but I thought it deserved a specific mention as posts are really of the same level as those on the Julia blog. They shouldn't be lost along all other blog posts.

jiahao:
:+1: for mentioning the Julia Computing blog.

Perhaps the wording could be improved so that it's clear that the company is not the sole source of "ongoing Julia development" (although no doubt an important source).

ViralBShah:
Good point - I think it should just say: See also the Julia Computing blog, and leave it that.

nalimilan:
I'd prefer keeping the explicit formulation about what kind of posts one can expect on that blog. What do you think of the new version?

ViralBShah:
I think its good.

PR #357, opened 2016-03-14
Update ideas-page.md, add QML bindings

ufechner7:
Add QML bindings for Julia. Should I add my email address here as possible mentor?


MikeInnes:
Looks good! Feel free to add an email but it sounds like you'll be keeping an eye on julia-users, in which case it's not a big deal.

PR #361, opened 2016-03-20
Update links for 0.4.5

tkelman:
Remove PPA because it is not currently being maintained, move Fedora links to below the generic Linux binaries.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -31,17 +31,13 @@ packages are pre-installed.
     <td colspan="6"> <a href="https://s3.amazonaws.com/julialang/bin/osx/x64/0.4/julia-0.4.5-osx10.7+.dmg">10.7+ 64-bit</a> </td>
 </tr>
 <tr>
-    <th> Ubuntu packages (.deb) </th>
-    <td colspan="6"> <a href="https://launchpad.net/~staticfloat/+archive/juliareleases">32/64-bit</a> </td>
+    <th> Generic Linux binaries </th>
+    <td colspan="3"> <a href="https://julialang.s3.amazonaws.com/bin/linux/x86/0.4/julia-0.4.5-linux-i686.tar.gz">32-bit</a> (<a href="https://julialang.s3.amazonaws.com/bin/linux/x86/0.4/julia-0.4.5-linux-i686.tar.gz.asc">GPG</a>)</td>
+    <td colspan="3"> <a href="https://julialang.s3.amazonaws.com/bin/linux/x64/0.4/julia-0.4.5-linux-x86_64.tar.gz">64-bit</a> (<a href="https://julialang.s3.amazonaws.com/bin/linux/x64/0.4/julia-0.4.5-linux-x86_64.tar.gz.asc">GPG</a>)</td>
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
Maybe the PPA needs not be dropped. I'm not trying to make more work for you, just pointing out that the PPAs seem to be working/maintained..* I know the PPAs are not strictly needed. Is this Saba guy the maintainer or was it you?

Just recently my update manager, showed 0.4.5, jumping over a version, I believe (this happened even after the EOL date, February 4, 2016, of Vivid.., after the update manager had been complaining since). Looking at launchpad:

julia 	0.4.5-wily1 	Elliot Saba (2016-04-06)
julia 	0.4.5-vivid1 	Elliot Saba (2016-04-06)
julia 	0.4.5-trusty1 	Elliot Saba (2016-04-06)
julia 	0.4.5-precise1 	Elliot Saba (2016-04-06) 

[I assume the PPA where used by downstream, e.g. Mint, and I'm not sure, upstream also (Debian has .deb, but also PPAs)?]

* http://releases.ubuntu.com/16.04/ Xenial Xerus

was just out the other day. I note it is missing from the list above. I'm not sure what I should read into that. I believe Ubuntu still supports PPAs in 16.04, while the Snappy format is new. Maybe PPAs are being phased out?

I think Julia could still use PPAs and for 16.04 you could use a PPA from an older Ubuntu version?

I haven't looked to much into the new Snappy format. Whether is would be a replacement or of use for Julia, or maybe more so for applications you would make with Julia only. Note however..:

http://www.zdnet.com/article/linux-expert-matthew-garrett-ubuntu-16-04s-new-snap-format-is-a-security-risk/

nalimilan:
@staticfloat What's the plan with the PPA?

staticfloat:
The PPA is low priority for me; as you can tell, I didn't even push out `0.4.4`, and instead skipped over to `0.4.5`.  I recently sent out a call to a few other people to see if they were interested in administering the PPA, but so far nobody has been.  Due to the simplicity of the generic linux binaries, this isn't a huge priority to me, even though I totally understand the desire to have Julia managed by a packaging system.

That Xenial is missing is simply due to the fact that I have not checked the "Xenial" box yet in my upload scripts.  I'm rectifying the issue now.

staticfloat:
Ironically, `Xenial` already has a Julia 0.4.5 package in it that it inherited from Debian.  That being said, the PPA is now up to date with `Xenial`.

nalimilan:
OK, great. It's unfortunate that we don't really know whether to recommend using the PPA or not. If you think you'll be able to keep it more or less up-to-date, maybe add it again to the downloads page?

staticfloat:
Without someone who is dedicated to keeping it up-to-date within a week of new releases, I think removing it from the downloads page is the right play.  If someone wants to step up and help me out with it, that's great, but unfortunately, I'm just too busy with other things right now to do it solo.  90% of the time, there are no issues and things work great, but every now and then, something goes screwy, and I have to spend a few hours tracking down what went wrong and how to fix it.

PR #363, opened 2016-03-22
We provide three ways, not two, to run Julia (correction to download page)

samuelpowell:


ViralBShah:
Can we make this "several"? This list keeps changing. :-)

samuelpowell:
@ViralBShah, done!

PR #366, opened 2016-03-28
Add blog post on challenging AbstractArray types

timholy:
As described [here](https://groups.google.com/d/msg/julia-dev/mKoVG54jygM/fnnwgn8nAAAJ), this is introductory material describing some challenges for supporting certain array types and operations. It will be paired with a shortly-to-be-filed GitHub issue outlining possible solutions.

If upon review you find this unsatisfying as a blog post, or find the split between problem & solution awkward, please recommend a better forum.

### DIFF
--- a/blog/_posts/2016-03-27-arrays-iteration.md
+++ b/blog/_posts/2016-03-27-arrays-iteration.md
@@ -1,0 +1,429 @@
+---
+layout: post
+title:  "Generalizing AbstractArrays: opportunities and challenges"
+author: <a href="http://holylab.wustl.edu">Tim Holy</a>
+---
+
+# Introduction: generic algorithms with AbstractArrays
+
+Somewhat unusually, this blog post is future-looking: it mostly
+focuses on things that don't yet exist. Its purpose is to lay out the
+background for community discussion about possible changes to the core
+API for `AbstractArray`s, and serves as background reading and
+reference material for a more focused "julep" (a julia enhancement
+proposal).  Here, often I'll use the shorthand "array" to mean
+`AbstractArray`, and use `Array` if I explicitly mean julia's concrete
+`Array` type.
+
+As the reader is likely aware, in julia it's possible to write
+algorithms for which one or more inputs are only assumed to be
+`AbstractArray`s.  This is "generic" code, meaning it should work
+(i.e., produce a correct result) on any specific concrete array type.
+In an ideal world---which julia approaches rather well in many
+cases---generality of code should not have a negative impact on its
+performance: a generic implementation should be approximately as fast
+as one restricted to specific array type(s).  This implies that
+generic algorithms should be written using lower-level operations that
+give good performance across a wide variety of array types.
+
+Providing efficient low-level operations is a different kind of design
+challenge than one experiences with programming languages that
+"vectorize" everything.  When successful, it promotes much greater
+reuse of code, because efficient, generic low-level parts allow you to
+write a wide variety of efficient, generic higher-level functions.
+
+Naturally, as the diversity of array types grows, the more careful we
+have to be about our abstractions for these low-level operations.
+
+# Examples of arrays
+
+In discussing general operations on arrays, it's useful to have a
+diverse collection of concrete arrays in mind.
+
+In core julia, some types we support fairly well are:
+
+- `Array`: the prototype for all arrays
+
+- `Range`s: a good example of what I often consider a "computed"
+  array, where essentially none of the values are stored in
+  memory. Since there is no storage, these are immutable containers:
+  you can't set values in individual slots.
+
+- `BitArray`s: arrays that can only store 0 or 1 (`false` or `true`),
+  and for which the internal storage is packed so that each entry
+  requires only one bit.
+
+- `SubArray`s: the problems this type introduced, and the resolution
+  we adopted, probably serves as the best model for the
+  generalizations considered here. Therefore, this case is discussed
+  in greater detail below.
+
+Another important class of array types in Base are sparse arrays:
+`SparseMatrixCSC` and `SparseVector`, as well as other sparse
+representations like `Diagonal`, `Bidiagonal`, and `Tridiagonal`.
+These are good examples of array types where access patterns deserve
+careful thought. Notably, despite many commonalities in "strategy"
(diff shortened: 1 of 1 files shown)


blog/_posts/2016-03-27-arrays-iteration.md
```diff
@@ -0,0 +1,427 @@
...
+
+But our triumph is short-lived. Let's return to the example of
+`copy!` above, and realize that `dest` and `src` might be two
+different array types, and therefore might be most-efficiently indexed
+with different iterator types.  We're tempted to write this as
+
+```jl
+function copy!(dest, src)
+    for (idest, isrc) in zip(eachindex(dest), eachindex(src))
+        dest[idest] = src[idest]
```
tkelman: is this intentionally not using `isrc`?

timholy: No, it's a typo. Thanks for catching.

blog/_posts/2016-03-27-arrays-iteration.md
```diff
@@ -0,0 +1,427 @@
...
+
+Unfortunately, I suspect that if we want to add support for certain
+new operations or types (specific examples below), it will force us to
+set the latter problem on fire.
+
+# Challenging examples
+
+Some possible new `AbstractArray` types pose novel challenges.
+
+## `ReshapedArrays` (#15449)
```
staticfloat: I suggest making any mention of a GitHub issue a link; there is the one in the title here, and two more in this section.

blog/_posts/2016-03-27-arrays-iteration.md
```diff
@@ -0,0 +1,427 @@
...
+In simple cases, there's an easy way to circumvent this performance
+problem: define a new iterator type that (internally) iterates over
+the parent `A`'s indexes directly.  In other words, create an iterator
+so that `B[I]` immediately expands to `A[I']`, and so that the latter
+has "ideal" performance.
+
+Unfortunately, this strategy runs into a lot of trouble when you need
+to keep two arrays in sync: if you want to adopt this strategy, you
+simply can't write `B[i,j]*v[j]` for matrix-vector multiplication
+anymore.  A potential way around *this* problem is define a new class
```
staticfloat: ...is **to** define a new class...

blog/_posts/2016-03-27-arrays-iteration.md
timholy: Good idea. I started writing this assuming the whole thing would be posted on GitHub, and then was appalled at how long it was.

blog/_posts/2016-03-27-arrays-iteration.md
```diff
@@ -0,0 +1,425 @@
...
+- `AbstractArray`s are specialized associative containers, in that the
+  allowable "keys" may be restricted by more than just their julia
+  type.  Specifically, the allowable keys must be representable as a
+  cartesian product of one-dimensional lists of values.  The allowed
+  keys may depend not just on the array type but also the specific
+  array (e.g., its size).  Attempted access by keys that cannot be
+  converted to one of the allowed keys, for that specific array,
+  result in `BoundsError`s.
+
+- For any given array, at least one complete independent set of keys
```
btracey: How does this interact with the Interpolations array that you mentioned in the intro? It has to be able to generate a complete set of floating point numbers?

jrevels:
Might be a config issue on my end, but I get an error when trying to build the site with this post:

```
➜  julialang.github.com git:(teh/arrays) jekyll serve
Configuration file: /Users/jarrettrevels/data/code/julialang.github.com/_config.yml
            Source: /Users/jarrettrevels/data/code/julialang.github.com
       Destination: /Users/jarrettrevels/data/code/julialang.github.com/_site
      Generating...
Error reading file /Users/jarrettrevels/data/code/julialang.github.com/blog/_posts/2016-03-27-arrays-iteration.md: (<unknown>): mapping values are not allowed in this context at line 3 column 36
```

The rest of the site is unaffected, but the formatting of this post becomes wonky:

screen shot 2016-03-29 at 10 57 57 am

timholy:
Maybe the last commit will fix it, @jrevels? Thanks for checking!

blog/_posts/2016-03-27-arrays-iteration.md
timholy: Hmm, good point. I guess by "complete" I'm back to thinking about discrete (integer-like) indexes. There have been arguments made that Interpolations should not be arrays and not use `getindex` to calculate values; one reaction is that this argument just got a little stronger.

Alternatively, we could discard this point. My intent here was to specify that, unlike traditional associative containers, the array doesn't _have_ to store the "keys."

tlycken: The "complete set" could also here be defined not by an actual list of values, but rather of intervals (just as today); an  `d`th index `i` is in the set if `1 <= i <= size(itp,d)`. This isn't conceptually much different from how a range object behaves: rather than storing all the values for all the allowed keys, there is an exact definition of which keys are allowed, and the values for those keys might be computed at indexing-time. (Interpolation arrays, like ranges, are also read-only.)

Arguably, though, iteration over interpolation arrays doesn't make much sense. The least weird thing to do (and what Interpolations.jl does today, to my knowledge) is to just return the values on each provided data point. In other words, iterating over an interpolation array yields the same results as iterating over the underlying data set (except that you can't set the value).

blog/_posts/2016-03-27-arrays-iteration.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: post
 title:  Generalizing AbstractArrays: opportunities and challenges
```
jrevels: Removing the web link didn't seem to fix it build issue - I think it's the unescaped second colon in `title: Generalizing AbstractArrays: opportunities and challenges` that's causing the trouble. Quoting the title fixed it for me, e.g. `title:  "Generalizing AbstractArrays: opportunities and challenges"`.

timholy:
OK, comments addressed. The reviews have been very helpful!

If no more concerns arise, I'll squash and merge this evening (or so).

tlycken:
For reference and easy navigation: JuliaLang/julia#15648 is the issue where discussion continues.

PR #379, opened 2016-05-10
Update 2012-03-11-shelling-out-sucks.md

jiahao:
Fix broken link to Part 2


ViralBShah:
Yes this has happened on many URLs, perhaps due to a jekyll upgrade or something.

PR #380, opened 2016-05-14
Add "Julia High Performance" to Books

aviks:
Disclaimer: I am the author.


ViralBShah:
I would add it as the first book. They are in the newest books first order. If you don't mind, can you add Malcolm's book too?

aviks:
Changed the order. Malcolm's book is already there.

aviks:
Thanks Viral.

PR #381, opened 2016-05-27
Add Julia Math Github group

oxinabox:
After Post: https://github.com/JuliaLang/julia/issues/16357#issuecomment-222144998
See Roadmap: https://github.com/JuliaMath/Roadmap.jl/issues/2

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -43,6 +43,7 @@ Various Julia projects are hosted under the following umbrella organizations on 
 * [JuliaCI](https://github.com/JuliaCI) – Continuous Integration Support for Julia packages
 * [JuliaDSP](https://github.com/JuliaDSP) – Digital signal processing
 * [JuliaGraphs](https://github.com/JuliaGraphs) – Graph Theory and Implementation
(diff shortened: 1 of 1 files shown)


community/index.md
```diff
@@ -43,6 +43,7 @@ Various Julia projects are hosted under the following umbrella organizations on
 * [JuliaCI](https://github.com/JuliaCI) – Continuous Integration Support for Julia packages
 * [JuliaDSP](https://github.com/JuliaDSP) – Digital signal processing
 * [JuliaGraphs](https://github.com/JuliaGraphs) – Graph Theory and Implementation
+* [JuliaMath](https://github.com/JuliaMath) -.Mathematics made easy in Julia
```
tkelman: extra . here

oxinabox: Thanks for catching this.
Its not actually an extra `.` causing a problem, so much as a `-` vs `–`
(I think hyphen vs en-dash) 
(odd that it is rendering differently the the coder in you line comment, or is it? Everyline has a `–.` in the git editor that i can see, but not in the diff...)

Issue #383, opened 2016-06-04
JuliaIO and others? missing from organizations page


TotalVerb:
@tkelman created some new organizations to lessen the load on JuliaLang's Travis. I don't know the full list of such organizations, but they should be added to the [community/index.md](https://github.com/JuliaLang/julialang.github.com/blob/master/community/index.md) file. At least https://github.com/JuliaIO seems to be missing, and probably more.

tkelman:
I think the only one that I personally created that isn't still empty is JuliaMath. I have a few more JuliaPackages, JuliaEditorSupport that are still placeholders at the moment.

ViralBShah:
submit PRs?

TotalVerb:
@ViralBShah see #384

oxinabox:
Every mentioned Org has now been added.
This can be closed

PR #384, opened 2016-06-04
Add JuliaIO organization

TotalVerb:
Part of #383. I think that issue should be left open as a reminder in case the new organizations get used.

Also fixes an en dash in BioJulia.


TotalVerb:
Oops, "package" is not the right term. Will fix commit message.

PR #395, opened 2016-07-16
The <col> element has no end tag.

krysros:


jiahao:
Confirmed fix locally.

I merged it on the command line (8e7a6ca2a2c7ddf37e4979d1330b21c5b9550962) but it's not registering here.

PR #403, opened 2016-08-27
Stop supporting 0.3.x

PallHaraldsson:
[skip ci]


PallHaraldsson:
Julia packages supporting 0.3.x seem to be dropping like flies. Drop support now, or no later than release of 0.5? It's not like 0.3.x has seen an update lately.. I don't know, do you want one final bug fix release before committing this?

ViralBShah:
This should not be merged until 0.5 is released. Also it is perhaps better to let @tkelman update these pages as part of the release process.

PR #404, opened 2016-08-27
Stop supportiing 0.3.x

PallHaraldsson:
[skip ci]


yuyichao:
I'll leave https://github.com/JuliaLang/julialang.github.com/pull/403 for Tony to close but at least we don't need two duplicated PRs on this.

PR #405, opened 2016-08-28
Fix markdown in `Graft` blog.

pranavtbhat:


tkelman:
you should always use a branch that isn't master of your fork when you make pull requests.

PR #407, opened 2016-09-05
Added SwitchTimeOpt.jl publication

bstellato:
I have added the paper "Second-Order Switching Time Optimization for Switched Dynamical Systems" describing SwitchTimeOpt.jl package to http://julialang.org/publications/


ViralBShah:
Is the diff so big because of whitespace issues?

bstellato:
I guess so. I have just added the paper to `publications/julia.bib` and `publications/_EDIT_ME_index.md`. Then, I generated the page with pandoc. 

Could it also be related to the new numbering of the other papers? I have added the paper at the end of the subsection "Numerical Optimization and Operations Research".

PR #411, opened 2016-09-13
Add Power nightlies link

tkelman:
Separate arm and power on a different table row from x86, since they aren't really "generic linux binaries" in the same sense as we build for x86. They won't work on any distribution other than the one the buildbot happens to be using.

cc @viralbshah


ViralBShah:
For arm and power, should we mention what distros they were built on and a warning that they may not work elsewhere if it is too different?

tkelman:
sure, where should the warning go? not much room for footnotes in the table

ViralBShah:
Just above or below the table?

ViralBShah:
Merging for now. Can clarify later.

PR #412, opened 2016-09-14
Add Julia for Data Science

scls19fr:


ViralBShah:
Best to put it in the released category, since this is going to be available in a few days.

scls19fr:
Ok. I've just moved it to the released category.

PR #417, opened 2016-09-20
Move 0.4.7 to old releases, update links to 0.5.0 final

tkelman:
and close #403

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -3,7 +3,7 @@ layout: default
 title:  Julia Downloads
 ---
 
-# Current Release (v0.4.7)
+# Current Release (v0.5.0)
 
 We provide several ways for you to run Julia:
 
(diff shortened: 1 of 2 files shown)


tkelman:
@ViralBShah want to double check?

ViralBShah:
I think the Juno IDE section is redundant, since we already mention Juno way up.

tkelman:
can delete that separately as it's not really related?

tkelman:
I also need to rewrite the Linux section of the platform-specific page to stop talking about the PPA, but that isn't this PR.

ViralBShah:
I tried it out on jekyll and looks great. What about the older releases page where 0.4 ought to be there?

tkelman:
What about it? I don't have jekyll, but did add the 0.4 table to oldreleases https://github.com/JuliaLang/julialang.github.com/pull/417/files#diff-935e9b612306b9717d6bdd7199655ba0

ViralBShah:
Hmm. I see it in the PR, but not in my jekyll.

ViralBShah:
Ah well - the URLs are absolute, so it was going to julialang.org. Tried the link manually and it is fine.

Not relevant here directly, but will 0.3 remain at critical bugfixes, or should it be unmaintained now and 0.4 go to critical bugfixes?

ViralBShah:
Good to merge.

ViralBShah:
I tried the os x, arm and power binaries. All looking good.

tkelman:
0.3 would be pretty difficult to make any more backport releases for, since the buildbot configuration has changed so much. Someone also deleted the "backport pending 0.3" label from the Julia issue tracker (please never do that again with a label used for tracking, if whoever did that sees this!) so anything that may have had it listed isn't really trackable any more.

0.4 I think we can continue maintaining for a while longer if we really need to. It'll take some time for everyone to migrate to 0.5, and there might still be a few things worth fixing on 0.4 anywhere it's easy to do.

ViralBShah:
In that case, should old releases just say unmaintained for 0.3 now? I would prefer to do that.

PR #423, opened 2016-09-26
Declare 0.3 unmaintained on oldreleases page

martinholters:
To be consistent with statement on downloads/index (ref #417).


ViralBShah:
I was meaning to do this myself. Thanks!

PR #427, opened 2016-09-26
RFC: Add StructuredQueries.jl summer project blog post

davidagold:
This PR adds my blog post describing my work on the jplyr.jl/StructureQueries.jl package this summer. It is admittedly a bit long; I hoped to include enough information to give helpful context to anybody (including my future self) who wants to contribute to the project. I'm happy to break it up into a series of blog posts, if that'd be more appropriate.

TO-DO:
- [x] tag StructuredQueries v0.0.2
- [x] tag AbstractTables v0.0.1
- [x] tag TablesDemo v0.0.1

### DIFF
--- a/blog/_posts/2016-10-03-StructuredQueries.md
+++ b/blog/_posts/2016-10-03-StructuredQueries.md
@@ -1,0 +1,390 @@
+## StructuredQueries.jl
+
+This post describes my work conducted this summer at the [Julia Lab](http://julia.mit.edu/) to develop [StructuredQueries.jl](https://github.com/davidagold/StructuredQueries.jl/), a generic data manipulation framework for [Julia](http://julialang.org/).
+
+Our initial vision for this work was much inspired by Hadley Wickham's [dplyr](https://github.com/hadley/dplyr) R package, which provides data manipulation verbs that are generic over in-memory R tabular data structures and SQL databases, and [DataFramesMeta](https://github.com/JuliaStats/DataFramesMeta.jl) (begun by [Tom Short](https://github.com/tshort)), which provides metaprogramming facilities for working with Julia `DataFrame`s.
+
+While a generic querying interface is a worthwhile end in itself (and has been discussed [elsewhere](https://groups.google.com/d/topic/julia-dev/jL2FSL4EneE/discussion)), it may also be useful for solving problems specific to in-memory Julia tabular data structures. We will discuss how a query interface suggests solutions to two important problems facing the development of tabular data structures in Julia: the *column-indexing* and *nullable semantics* problems. So, the present post will describe both the progress of my work and also discuss a wider scope of issues concerning support for tabular data structures in Julia. I will provide some context for these issues; the reader should feel free to skip over any uninteresting details.
+
+---
+
+Recall that the primary shortcoming of [DataArrays.jl](https://travis-ci.org/JuliaStats/DataArrays.jl) is that it does not allow for type-inferable indexing. That is, the means by which missing values are represented in `DataArray`s -- i.e. with a token `NA::NAtype` object -- entails that the most specific return type inferable from `Base.getindex(df::DataArray{T}, i)` is `Union{T, NAtype}`. This means that until Julia's compiler can better handle small `Union` types, code that naively indexes into a `DataArray` will perform unnecessarily poorly.
+
+[NullableArrays.jl](https://github.com/JuliaStats/NullableArrays.jl) [remedied](http://julialang.org/blog/2015/10/nullablearrays) this shortcoming by representing both missing and present values of type `T` as objects of type `Nullable{T}`. However, this solution has limitations in other respects. First, use of `NullableArray`s does nothing to support type inference in column-indexing of `DataFrame`s. That is, the return type of `Base.getindex(df::DataFrame, field::Symbol)` is not straightforwardly inferable, even if `DataFrame`s are built over `NullableArray`s. Call this first problem the *column-indexing problem*. Second, NullableArrays introduces certain difficulties centered around the `Nullable` type. Call this second problem the *nullable semantics problem*.
+
+The column-indexing problem is [well-documented](http://www.johnmyleswhite.com/notebook/2015/11/28/why-julias-dataframes-are-still-slow/). To see the difficulty, consider the following function
+```
+function f(df::DataFrame)
+    A = df[:A]
+    x = zero(eltype(A))
+    for i in eachindex(A)
+        x += A[i]
+    end
+    return x
+end
+```
+where `df[:A]` retrieves the column named `:A` from `df`. A user might reasonably expect the above to be idiomatic Julia: the work is written in a `for` loop that is wrapped inside a function. However, this code will not be (ahead-of-time) compiled to efficient machine instructions because the type of the object that `df[:A]` returns is cannot be inferred during static analysis. This is because there is nothing the `DataFrame` type can do to communicate the `eltype`s of its columns to the compiler.
+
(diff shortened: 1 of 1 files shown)


blog/_posts/2016-09-17-StructuredQueries.md
```diff
@@ -0,0 +1,386 @@
...
+│ 5   │ 10.0               │
+│ 6   │ 10.8               │
+│ 7   │ 9.2                │
+│ 8   │ 10.0               │
+│ 9   │ 8.8                │
+│ 10  │ 9.8                │
+⋮
+with 140 more rows.
+```
+Whatever the name of the dummy source (minus the `:`) was in the query must be the key in the kwarg passed to `collect`:
```
nalimilan: "Else, the collection will fail' or something similar?

blog/_posts/2016-09-17-StructuredQueries.md
```diff
@@ -0,0 +1,386 @@
...
+    end
+end
+```
+with natural analogues for methods with n-ary arguments. This process is a bit cumbersome, but it would not be difficult to automate with a macro with which one could annotate the original definition `f(x::T)`.
+
+I don't think this solution is satisfactory. One must somehow decide which functions should be lifted in this manner. Furthermore, if one cannot edit the definition of a function then a macro is of no use; one must manually introduce the lifted variant.
+
+There is a further problem. If one wants to support lifting over arguments with "mixed" signatures -- i.e. signatures in which some argument types are `Nullable` and some are not -- then one has either to extend the promotion machinery or to define methods for mixed signatures, e.g. `+{T}(x, y::Nullable{T})`. That may end up being a lot of methods. Even if their definition can be automated with metaprogramming, the compilation costs associated with method proliferation may be considerable (but I haven't tested this).
+
+Finally, there is the problem described in [NullableArrays.jl#148](https://github.com/JuliaStats/NullableArrays.jl/issues/148#issuecomment-249335994).
```
nalimilan: Say a word about it! :-)

blog/_posts/2016-09-17-StructuredQueries.md
```diff
@@ -0,0 +1,386 @@
...
+
+I think that a better way to implement standard lifting semantics is by means of a higher-order function -- that is, on 0.5 where higher-order functions are performant. Such a function -- call it `lift` -- might look like the following:
```
nalimilan: "on Julia 0.5"

blog/_posts/2016-09-17-StructuredQueries.md
```diff
@@ -0,0 +1,386 @@
...
+function lift(f, x)
+    if hasvalue(x)
+        return Nullable(f(x))
+    else
+        U = Core.Inference.return_type(f, (typeof(X),))
+        return Nullable{U}()
+    end
+end
+```
+This definition can naturally be extended to methods with more than one argument. The primary advantage of this approach over lifting by method extension is its generality: one needs only to define the  Note that as long as `hasvalue` has some generic fallback method for non-`Nullable` arguments, this one definition covers both standard and mixed-signature lifting. (Ideally one would ensure that the code is optimized for when types are non-`Nullable`; in particular, one would ensure that the dead branch is removed -- cf. [julia#18484](https://github.com/JuliaLang/julia/pull/18484).)
```
nalimilan: Bug in the sentence.

blog/_posts/2016-09-17-StructuredQueries.md
```diff
@@ -0,0 +1,386 @@
...
+
+Recall that the present query framework extracts the "value expression" of a query argument (for instance, `B * C` in the query argument `C = A * B`) and generates a lambda that mimics the former's structure (in this case, `row -> row[1] * row[2]`). A proposed modification (see [AbstractTables#2](https://github.com/davidagold/_AbstractTables.jl/issues/2)) to this process is to modify the AST of the value expression (`A * B`) by appropriately inserting calls to `lift`, e.g.
+```
+row -> lift(*, row[1], row[2])
+```
+While there is a [simpler way](https://github.com/davidagold/AbstractTables.jl/blob/2a7771ce865b961fa0e454508ce8b7aa6a85e1fd/src/column_indexable/query/select.jl#L43-L48) to achieve standard lifting semantics, this approach (which is currently employed by the column-indexing collection machinery) does not easily support non-standard lifting semantics such as 3VL.
```
nalimilan: Spell 3VL in full.

blog/_posts/2016-09-17-StructuredQueries.md
```diff
@@ -0,0 +1,386 @@
...
+
+Going further, one could imagine employing `-> ...` as syntactic sugar for `|> select(...)`, so that the above may be expressed as
+```
+@collect i in X -> i & false
+```
+which seems an appropriate combination of a lambda declaration and a generator declaration -- appropriate because the result of such a command would, under the proposed desugaring, essentially resolve to `map(i -> i & false, X)` but in which the lambda `i -> i & false` may be given a wider variety of lifting semantics since the macro `@collect` has access to the syntax tree defining said lambda. Such syntactic sugar may also make transformations of tabular data structures more terse, since attributes needn't be declared in such cases. For instance:
+```
+@collect iris -> *, twice_sepal_length = 2 * sepal_length
+```
+(where `select(*, ...)` is equivalent to horizontally concatenating (from the right) the table produced by applying the transformations in `...` to the original table).
```
nalimilan: `select(*, ...)` does not appear in the expression.

davidagold:
Thank you for the comments, @nalimilan. Do you think it should be split into multiple posts?

ViralBShah:
Length is not an issue at all. In fact it is nice to have some long form posts in this age of 140 characters.

ViralBShah:
@davidagold Please ping once the review is addressed. Also pinging @johnmyleswhite here. We can merge soon.

ViralBShah:
Need to remember to change the date.

nalimilan:
Yes, the length isn't a problem. Better have everything in one place.

davidagold:
@ViralBShah I have incorporated Milan's edits and revised some sections. I think this is ready to go.

davidagold:
Thanks for the merge. Am I the only person for whom this is not displaying properly? http://julialang.org/blog/2016/10/StructuredQueries

ViralBShah:
Just fixed it.

ViralBShah:
Something still amiss in the example in Dummy Sources. Any idea how to fix it?

davidagold:
There are a lot of places where code blocks aren't being properly displayed. I'll take a look.

blog/_posts/2016-10-03-StructuredQueries.md
```diff
@@ -0,0 +1,390 @@
...
+end
+```
+As long as `f_inner` does not get inlined, type inference will run "at" the point at which the body of `f` calls `f_inner` and will have access to the `eltype` of `df[:A]`, since the latter is passed as an argument to `f_inner`.
+
+This strategy of introducing a function barrier also works when one requires multiple columns. For instance, suppose I wanted to generate a new column `C` where `C[i] = g(A[i], B[i])`. The following solution is type-inferable since the type parameters of the zipped iterator `zip(A, B)` reflects the `eltype`s of `A` and `B`:
+
+```
+function f(g, df)
+    A, B = df[:A], df[:B]
+    C = similar(A)
```
quinnj: It probably works in this case, but I would imagine that calling `C = similar(A)` would not work generally since you effectively need to do `C = similar(Core.Inference.return_type(g, (A[1], B[1]))`. Probably not a big deal thoug

blog/_posts/2016-10-03-StructuredQueries.md
```diff
@@ -0,0 +1,390 @@
...
+
+The difficulties associated with method extension lifting are not insurmountable, but the solution -- namely, keeping a repository of lifted methods -- requires an undetermined amount of maintenance and coordination.
+
+Another way to implement standard lifting semantics is by means of a higher-order function -- that is, on Julia 0.5 where higher-order functions are performant. Such a function -- call it `lift` -- might look like the following:
+```
+function lift(f, x)
+    if hasvalue(x)
+        return Nullable(f(x))
+    else
+        U = Core.Inference.return_type(f, (typeof(X),))
```
quinnj: Should be a lowercase `x` here.

PR #429, opened 2016-09-27
Updated with publication RSLLBHGL16

roding:
Updated with publication RSLLBHGL16. Have not run Jekyll, asking for help with that, thanks,


jiahao:
List of publications updated in 5314ca5

PR #431, opened 2016-09-30
fix brazilian meetup

felipenoris:
Rio de Janeiro meetup no longer exists.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -132,6 +132,8 @@ South America
 
 * [São Paulo Julia Meetup](http://www.meetup.com/Sao-Paulo-Julia-Meetup/)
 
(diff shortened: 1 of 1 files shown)


Ismael-VC:
I still see the meetup online. What do you mean it no longer exists? Has it been inactive? Have you tried contacting the organizer? Consider why not revive it yourself? I'm sure you can ask the meetup staff to intervene if the original organizer has vanished, to name you or some one else who is interested in becoming an organizer in his place.

I'll share this at julia-i18n and JuliaLangPt.

Ismael-VC:
False alarm, I understood that there meetup was inactive, or something like that, anyway, if there is something I can do to help the Portuguese community, just give me a call!

felipenoris:
Just started a new Julia Rio de Janeiro meetup.

PR #432, opened 2016-10-02
Make benchmark table not spillover on home page

djsegal:
With this change the `page-width` is properly `100%`

### DIFF
--- a/css/screen.css
+++ b/css/screen.css
@@ -155,6 +155,10 @@ hr {
 
 /* benchmarks table */
 
+div.cs-benchmark-table {
+  overflow-x: scroll;
+}
+
 table.benchmarks {
   margin-left: auto;
   margin-right: auto;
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Much better, thanks!

ViralBShah:
Thanks for this! Off-topic, but I should point out that we really should have the plot on this page instead of the table. Will do this at some point in a julialang website revamp.

Ismael-VC:
This closes #254 

I also think we should have the plot and table switch places, so the plot is on the homepage instead, but it would be cool if it's also ordered from smallest to largest run time:

* https://github.com/JuliaLang/julialang.github.com/issues/254#issuecomment-124532904

> I'm all for that change. I'd also really like to sort the table differently – it used to be from smallest average run time to largest, which is far more helpful and informative that alphabetical order.

PR #439, opened 2016-10-11
fix filenames

felipenoris:


felipenoris:
Page was not rendering word `<version>`

PR #441, opened 2016-10-12
Modify title

ranjanan:
I wonder if "on" is the right preposition here? Also, another possibly useful title would be "Looking Back at JuliaCon 2016"


ViralBShah:
Either is ok. Let me know when you update it and I will merge. :-)

PR #447, opened 2016-10-21
Update ARM nightly link and rename it to ARMv7

yuyichao:
Remove "32bits" since that's implied by v7.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -36,7 +36,7 @@ JuliaBox, all of these plotting packages are pre-installed.
 </tr>
 <tr>
     <th> Linux builds for other architectures </th>
-    <td colspan="3"> <a href="https://julialang.s3.amazonaws.com/bin/linux/arm/0.5/julia-0.5.0-linux-arm.tar.gz">ARM 32-bit hard float</a> (<a href="https://julialang.s3.amazonaws.com/bin/linux/arm/0.5/julia-0.5.0-linux-arm.tar.gz.asc">GPG</a>)</td>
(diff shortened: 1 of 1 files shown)


tkelman:
I don't think everyone is going to know that, best to leave the 32 bit there.

yuyichao:
Everyone that want to use the binary should know it. The v7 is also much more important than the 32bit.

tkelman:
I'm in a room with a few dozen people who use these binaries every day, and in terms of what they're familiar with it's the exact opposite.

tkelman:
will this also change binaries manually built from release-0.5? if so, then `contrib/prepare_release.sh` in the Julia repo will need to be updated (and backported) as well

yuyichao:
Not sure what exactly you mean by "manually built from release-0.5" but the change is caused by using a newer gcc that's compiled for an armv7 triple. If you build any version manually using a compiler that gives you `arm-***` triple the name would be the same with what it was before.

tkelman:
I mean the binaries built on the buildbot, including release binaries of forthcoming 0.5.x backports.

yuyichao:
Then yes. Unless the compiler on the buildbot is recompiled with a different triple the new binaries will have armv7l in them by default.

yuyichao:
Actually the name of the binary on http://julianightlies.s3.amazonaws.com/ didn't change so `prepare_release.sh` does not need to be changed. We'll want to differentiate them if/when we have v6/v8 binaries but that can happen later.

ViralBShah:
Good to have the 32-bit in there. I didn't know that v7 was 32-bit only.

PR #448, opened 2016-10-22
Added Julia Cookbook

Dawny33:
- Added the Julia Cookbook, along with the details of the book  (Publisher - Packt)
- I am the author of the book.


ViralBShah:
Great to have the book here.

PR #456, opened 2016-11-07
Add GBZWP16 publication

carlobaldassi:
New publication. I built it locally and everything seems to be working fine.
Note: after building I had some changes in `Gemfile.lock`. I haven't included those. Should I?


ViralBShah:
Merging for now. Why is Gemfile.lock checked in? Seems like it shouldn't be from the sound of it.

carlobaldassi:
> Why is Gemfile.lock checked in? Seems like it shouldn't be from the sound of it.

That's what I was thinking also.

PR #457, opened 2016-11-14
add link to discourse page

simonbyrne:


ViralBShah:
Cc @vchuravy @StefanKarpinski

StefanKarpinski:
Sure, although it might be a little early since we haven't moved julia-users over yet.

simonbyrne:
I've purposely left julia-users there, with the intention that it be removed when the final move is made. It seems odd though to leave julia-dev and julia-stats up.

StefanKarpinski:
Yeah, fair enough.

PR #465, opened 2016-12-13
Add GGPlots.jl project

ChrisRackauckas:
GGPlots.jl adds a grammar of graphics API to Plots.jl. @tbreloff already has some starter code though there's still more to do, making it an easy project to get started with. But given chats with people throughout the ecosystem, many people are holdouts on Plots.jl until it has this style of API (I guess a lot of data scientists really like it?) so it would be a very big gain for Jullia.


ChrisRackauckas:
I added a few links to the Plots.jl section.

PR #466, opened 2016-12-13
Add to random number generator project

ChrisRackauckas:
@sunoru 's RNG.jl project from last year was really great, but you can see there are some remaining issues. It would be nice to get another hand on it next year. Also, I think RNG.jl should be migrated to JuliaMath so that way it can start becoming a more standard part of Julia, maintained by more members of the community.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -188,14 +188,14 @@ This project proposal is to implement a native Julia framework for distributed e
 
 [Monte Carlo methods](https://en.wikipedia.org/wiki/Monte_Carlo_method) are becoming increasingly important in large-scale numerical computations, requiring large quantities of random numbers.
 To ensure accuracy of the simulated systems, it is critical that the [pseudorandom number generator](https://en.wikipedia.org/wiki/Pseudorandom_number_generator) is both fast and reliable, avoiding problems with periodicity and dependence, robust to statistical tests such as the [Crush suite](https://github.com/andreasnoack/RNGTest.jl).
-Challenges are even greater in massively parallel computations, which require going beyond running many copies of serial algorithms for generating pseudorandom numbers, due to well-known synchronization effects which can compromise the quality and uniformity of of random sampling.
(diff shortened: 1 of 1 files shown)


ViralBShah:
I am not sure why discuss the package transfer here. Perhaps should be discussed in the relevant packages. It is a good suggestion. However, the owner org may not want it without a clear maintainer.

sunoru:
I agree. Actually I really want more people to join and discuss the issues..

ChrisRackauckas:
>Actually I really want more people to join and discuss the issues..

I think that's just because it's much less visible than it would be if it was in the JuliaMath org. As per @ViralBShah 's request, I made a new issue in the JuliaMath Roadmap repo to discuss this:

https://github.com/JuliaMath/Roadmap.jl/issues/5

soc/ideas-page.md
```diff
@@ -188,14 +188,14 @@ This project proposal is to implement a native Julia framework for distributed e
...
 To ensure accuracy of the simulated systems, it is critical that the [pseudorandom number generator](https://en.wikipedia.org/wiki/Pseudorandom_number_generator) is both fast and reliable, avoiding problems with periodicity and dependence, robust to statistical tests such as the [Crush suite](https://github.com/andreasnoack/RNGTest.jl).
-Challenges are even greater in massively parallel computations, which require going beyond running many copies of serial algorithms for generating pseudorandom numbers, due to well-known synchronization effects which can compromise the quality and uniformity of of random sampling.
+Challenges are even greater in massively parallel computations, which require going beyond running many copies of serial algorithms for generating pseudorandom numbers, due to well-known synchronization effects which can compromise the quality and uniformity of of random sampling. A previous Google Summer of Code project created RNG.jl, and there are more random number generating algorithms to explore.
```
StefanKarpinski: Link to RNG.jl?

ChrisRackauckas:
Added the link.

ViralBShah:
@sunoru In that case, why not write a julialang blog post about what RNG.jl is and invite folks to join?

sunoru:
Oh you are right! I had tried to write a final post but to be honest I didn't put it in a high priority so I completely forgot it when the new semester began.... Perhaps I do need to finish it.

PR #467, opened 2016-12-13
Theme: Numerical Methods for Differential Equations

ChrisRackauckas:
I created a theme for numerical methods of differential equations. These are some projects I have been leaving around JuliaDiffEq which I think are both easily accessible to newcomers, don't require too much mathematical sophistication (undergraduate mathematics or graduate non-math like CS/Engineering would suffice), but also have very tangible and useful results. 

Many of these proposals give ecosystem wide results, for example the global sensitivity analysis results using the common interface would be able to be used with OrdinaryDiffEq.jl, ODE.jl, Sundials.jl, DASKR.jl, DASSL.jl, and I can keep listing. So these would have a very large affect on the ecosystem as a whole.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -111,7 +111,6 @@ seeking out opportunties for possible improvements along the way, such as suppor
 `Float32` and `BigFloat`, exploiting fused multiply-add operations, and improving errors
 and boundary cases.
 
-
 ## Matrix functions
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Note that post-Stefan's review, I have added links to the mentioned packages to match what was requested in the other PRs.

ChrisRackauckas:
I merged the Linear Algebra changes into this branch and then made the spacings throughout the page the same. I think that all of the topics related to data should be a theme as well, so I'll make that a separate PR.

PR #468, opened 2016-12-13
Native Usage of LinearMaps in Iterative Solvers

ChrisRackauckas:
As written:

>While one normally thinks of solving the linear equation Ax=b with A being a matrix, this concept is more generally applied to A being a linear map. In many domains of science, this idea of directly using a linear map instead of a matrix allows for one to solve the equation in a more efficient manner. Iterative methods for linear solving only require the ability compute `A*x` in order solve the system, and thus these methods can be extended to use more general linear maps. By restructuring IterativeSolvers.jl to use `LinearMap` types from LinearMaps.jl, these applications can be directly supported in the library.

I think this doesn't have much of a barrier to entry, but as multiple people have found this issue in IterativeSolvers.jl interesting, I think it would be a very nice GSoC project to finally implement this functionality.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -108,16 +108,6 @@ The user-facing API should have both a high-level, easy-to-use interface and a l
 
 **Expected Results**: A production-quality ODE/DAE solver package.
 
-
-
-## Native Julia implementations of iterative solvers for numerical linear algebra
-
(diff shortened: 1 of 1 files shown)


soc/ideas-page.md
```diff
@@ -329,9 +331,6 @@ Julia employs several techniques that are novel in the field of high performance
 
 # Theme: Julia Graphics and User Interfaces
 
-## 2D Graphics Improvements
```
tkelman: unrelated deletion?

ChrisRackauckas: Oh yes. My bad. I was thinking of putting in a PR for this. Didn't realize I had this on the same branch. Thank you catching this.

ChrisRackauckas:
I opened up a new PR here since this one seemed to have more issues than that (I wonder what happened...)

PR #469, opened 2016-12-13
Update ideas-page.md to include LightGraphs parallelism effort

sbromberger:
Added parallel LightGraphs to proposed GSoC themes.


sbromberger:
Thanks, @StefanKarpinski - I'm hoping we can do some really cool stuff with LightGraphs on HPC.

PR #470, opened 2016-12-13
Update linear algebra projects

ChrisRackauckas:
New version of https://github.com/JuliaLang/julialang.github.com/pull/468 since that PR seems to have been borked.

This creates a new theme around numerical linear algebra, moves the related projects there, and adds a new one based on Native Usage of LinearMaps in Iterative Solvers 

>While one normally thinks of solving the linear equation Ax=b with A being a matrix, this concept is more generally applied to A being a linear map. In many domains of science, this idea of directly using a linear map instead of a matrix allows for one to solve the equation in a more efficient manner. Iterative methods for linear solving only require the ability compute A*x in order solve the system, and thus these methods can be extended to use more general linear maps. By restructuring IterativeSolvers.jl to use LinearMap types from LinearMaps.jl, these applications can be directly supported in the library.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -90,34 +90,6 @@ While many tricks for presolving LPs are well known, there is significant room f
 imagination in writing a presolve for SDP; the project might well lead to a publication
 were the student so inclined.
 
-## PETSc integration for scalable technical computing
-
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This accidentally has an unrelated deletion for the old numerical differential equation solver part, which is supersceded anyways by https://github.com/JuliaLang/julialang.github.com/pull/467 . Sorry for the mess.

ViralBShah:
let's get this merged. rebase?

ChrisRackauckas:
Merged these changes underneath the LightGraphs project proposal.

PR #472, opened 2016-12-17
Update nightly text

ChrisRackauckas:
The proposal for this textual change come from this thread:

https://discourse.julialang.org/t/gurobi-jl-precompile-fails-after-upgrade/990

among many other where users have had issues using packages on nightlies, thinking that it's advisable or supported like that in some way. Thus I am making this wording stronger, even mentioning "You can expect many packages not to work with this version" since generally that is the case, and maybe that will help more users declare upfront that the problem is due to using the nightly builds.

It's still a little redundant so maybe it can be workshoped more.


ViralBShah:
Yes, this is better wording.

PR #474, opened 2016-12-18
Group GSoC Projects by themes

ChrisRackauckas:
This PR breaks up the giant " Scalability of technical computing for big data applications" into smaller and more mangagable themes. That main theme is still there for holding projects which did not fit in other places (and it makes sense: what's there is now SIMD, threading, and parallelism). 

Numerical optimization was pulled out to its own theme. I also made a theme which I called "Base Mathematics Libraries" for:

- Upgrading openlibm
- Special functions
- Matrix functions
- Fast bignums in Julia
- Random number generation

Maybe there could be a better name for this theme?

Also, the data related projects were moved to a theme "Data Handling".

Together, I think this makes it easier to navigate. Note that this is branched off of https://github.com/JuliaLang/julialang.github.com/pull/467, so it would require that first.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -111,7 +111,6 @@ seeking out opportunties for possible improvements along the way, such as suppor
 `Float32` and `BigFloat`, exploiting fused multiply-add operations, and improving errors
 and boundary cases.
 
-
 ## Matrix functions
 
 Matrix functions maps matrices onto other matrices, and can often be interpreted as generalizations of ordinary functions like sine and exponential, which map numbers to numbers. Once considered a niche province of numerical algorithms, matrix functions now appear routinely in applications to cryptography, aircraft design, nonlinear dynamics, and finance.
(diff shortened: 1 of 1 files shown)


ViralBShah:
Much of the openlibm work is being done by @musm and we should refer to that.

ViralBShah:
I can merge this when ready.

ChrisRackauckas:
Yes, I think the OpenLibm stuff should change to being about Amal.jl and @musm's work, but I am doing the PRs one at a time to hit different points. This one is about changing the groupings. Another will focus on updating the text for the OpenLibm parts.

Should potential mentors also be added (in a separate PR)?

ViralBShah:
Makes sense. These projects have accumulated over time, and when we get close to GSoC again, we can refresh for being current and potential mentors. Let's just do the reorganization for now.

ChrisRackauckas:
Fixed merge conflicts and ensured that the spacing/capitalizations used in different themes are the same.

PR #476, opened 2016-12-26
A new learning path added to the books page

packtpartner:


ViralBShah:
Please update the PR and mention that it is a combination and curation of the three separate books by the three authors. Also, has the code in all 3 books been harmonized to work with Julia 0.5? I suspect that the books came out at different times and for different versions of Julia.

packtpartner:
@ViralBShah Thanks for the reply. I will make the suggested changes

Issue #445, opened 2016-10-13
Missing Orgs


oxinabox:
From  #444 so as to keep that on topic of ordering.
- https://github.com/JuliaArrays
- https://github.com/JuliaGL

A hand full more, just from a quick look around:
- https://github.com/JuliaImages
- https://github.com/JuliaGeometry
- https://github.com/JuliaGraphics

(You know what these all have in common? @SimonDanisch )

---

Also 
https://github.com/JuliaText
which I think is dead https://github.com/JuliaText/Roadmap.jl/issues/4
(which is a problem for me, as that is my field. NLP/Textprocessing/ComputationalLingistics)

oxinabox:
JuliaML

ChrisRackauckas:
These were addressed by #477.

Issue #444, opened 2016-10-13
Orgs List is getting long


oxinabox:
Right now the Org list is has 25 list items.
(Including the one from my PR #443) 

It is not really sorted at all:
## Current

> - [JuliaLang](https://github.com/JuliaLang) – [The language](http://www.julialang.org/)
> - [JuliaStats](https://github.com/JuliaStats) – [Statistics](http://www.juliastats.org/)
> - [JuliaOpt](https://github.com/JuliaOpt) – [Optimization](http://www.juliaopt.org/)
> - [JuliaParallel](https://github.com/JuliaParallel) – [Parallel programming in Julia](http://juliaparallel.github.io/)
> - [JuliaDB](https://github.com/JuliaDB) – Various database drivers for Julia
> - [JuliaGPU](https://github.com/JuliaGPU) – GPU computing
> - [BioJulia](https://github.com/BioJulia) – Biology
> - [JuliaQuantum](https://github.com/JuliaQuantum) – [Julia libraries for quantum science and technology](http://juliaquantum.github.io/)
> - [JuliaAstro](https://github.com/JuliaAstro) – Astronomy
> - [JuliaQuant](https://github.com/JuliaQuant) – Finance
> - [JuliaSparse](https://github.com/JuliaSparse) – Sparse matrix solvers
> - [JuliaDiff](https://github.com/JuliaDiff/) – [Differentiation tools](http://www.juliadiff.org/)
> - [JuliaWeb](https://github.com/JuliaWeb) – Web stack
> - [JuliaCI](https://github.com/JuliaCI) – Continuous Integration Support for Julia packages
> - [JuliaDSP](https://github.com/JuliaDSP) – Digital signal processing
> - [JuliaGraphs](https://github.com/JuliaGraphs) – Graph Theory and Implementation
> - [JuliaMath](https://github.com/JuliaMath) – Mathematics made easy in Julia
> - [JuliaDiffEq](https://github.com/JuliaDiffEq) – Differential equation solving and analysis
> - [JuliaIO](https://github.com/JuliaIO) – IO-related functionality, such as serialization
> - [JuliaInterop](https://github.com/JuliaInterop) – Easy interoperability between Julia and not Julia
> - [JuliaLangEs](https://github.com/JuliaLangEs) – Julia resources in the Spanish language
> - [JuliaTime](https://github.com/JuliaTime) – Date and time libraries
[...]

PR #477, opened 2016-12-26
Add orgs and subdivide

ChrisRackauckas:
Implements #444 and #445. It adds the orgs of #445 and then subdivides as suggested in #444, using a consensus between the different methods mentioned in the thread. While you can always argue some changes in "what goes where", I think most of the categorizations here are well accepted enough to make this easier to read.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -38,34 +38,58 @@ There is also a list of [packages](http://pkg.julialang.org/) for Julia, many of
 Various Julia projects are hosted under the following umbrella organizations on GitHub:
 
 * [JuliaLang](https://github.com/JuliaLang) – [The language](http://www.julialang.org/)
-* [JuliaStats](https://github.com/JuliaStats) – [Statistics](http://www.juliastats.org/)
-* [JuliaOpt](https://github.com/JuliaOpt) – [Optimization](http://www.juliaopt.org/)
-* [JuliaParallel](https://github.com/JuliaParallel) – [Parallel programming in Julia](http://juliaparallel.github.io/)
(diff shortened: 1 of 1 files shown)


ViralBShah:
Yes, these are by and large good. Perhaps signal processing ought to be a section with JuliaImage and JuliaDSP in it. There is also an org or two for controls systems stuff.

ViralBShah:
I wouldn't change anything yet though.

community/index.md
```diff
@@ -38,34 +38,58 @@ There is also a list of [packages](http://pkg.julialang.org/) for Julia, many of
...
+## Computing
+
+* [JuliaArray](https://github.com/JuliaArrays)
+* [JuliaBerry](https://github.com/JuliaBerry) – [Julia resources for the Raspberry Pi](https://juliaberry.github.io/)
+* [JuliaCI](https://github.com/JuliaCI) – Continuous Integration Support for Julia packages
+* [JuliaGPU](https://github.com/JuliaGPU) – GPU computing
+* [JuliaInterop](https://github.com/JuliaInterop) – Easy interoperability between Julia and not Julia
+* [JuliaIO](https://github.com/JuliaIO) – IO-related functionality, such as serialization
 * [JuliaParallel](https://github.com/JuliaParallel) – [Parallel programming in Julia](http://juliaparallel.github.io/)
+* [JuliaPolyhedra](https://github.com/JuliaPolyhedra) – [Polyhedral computation](https://juliapolyhedra.github.io/)
```
tkelman: this isn't the right category for this

ChrisRackauckas: Move it over to math?

tkelman: yes, it's more or less a special class of computational geometry

PR #478, opened 2016-12-26
add Plots to downloads page

ChrisRackauckas:
I think it's about time we added Plots.jl to the downloads page. I tried to keep simple while mentioning Plots' two main features: backend-independence and recipes. The code example highlights the first (and ease of use), while a link to the ecosystem page highlights the extension functionality.

### DIFF
--- a/downloads/plotting.md
+++ b/downloads/plotting.md
@@ -7,6 +7,29 @@ title:  Plotting in Julia
 
 Plotting in Julia is available through external packages.
 
+## Plots
+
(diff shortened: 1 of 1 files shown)


ViralBShah:
Shouldn't it install the plotly package too? Or does Plots.jl have a dependency on plotly?

ChrisRackauckas:
Plotly only needs its Javascript, so Plots.jl installs the Javascript to use Plotly on its build.

PR #480, opened 2016-12-28
blog post on syntactic loop fusion in Julia 0.6

stevengj:
Since I commonly get questions on this, I thought it would be useful to have a blog post on the upcoming "syntactic loop fusion" (i.e. dot calls) feature of Julia 0.6.

cc: @Sacha0, @yuyichao, @pabloferz


blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,449 @@
...
+a lot of memory (more than an order of magnitude!)
+
+If the array `X` is small, then the performance cost of allocating
+these temporary arrays is significant: heap allocation is expensive.
+If the array `X` is large enough, then the time for the heap allocation
+and garbage collection may be negligible, but you pay a *different* performance price
+from the fact that you have 12 loops (12 passes over memory) compared
+to one, largely because of the loss of [cache locality](https://en.wikipedia.org/wiki/Locality_of_reference).
+
+In particular, main computer memory (RAM) is relatively slow (much slower than
```
johnmyleswhite: I think if I didn't already feel comfortable with this material, this section would confuse me badly because I'd assume that temporal locality would only benefit me if the cost of computing all of my fused operations on `X[i]` were small relative to the cost of loading `X[i]` from main memory into the cache. But it might not have been obvious to me that adding two numbers together is so much less expensive than loading them into cache.

stevengj: RAM latency is about 15ns these days, no?  This is an eternity compared to the cost of adding two numbers.  Of course, because you are typically reading RAM consecutively, it's not quite so bad, but it's still a huge slowdown to have a cache miss for every arithmetic operation.

stevengj: Well, I guess a cache miss for every 8 arithmetic operations on doubles, assuming 64-byte cache lines.

stevengj: Is there a way to make it clearer while remaining brief?

pabloferz:
Maybe it's worth mentioning that both at a syntactic level and thanks to the implementation of `broadcast`, it is possible to handle mixtures of scalars, tuples and/or arrays? I believe that is also non-common in other programming languages.

ViralBShah:
This is a great post. Just a thought - can timings be included in the post showing the differences in performance for the approaches - or perhaps post a notebook along that users can play with.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,449 @@
...
+code](https://en.wikipedia.org/wiki/Machine_code) for the function is (in Julia,
+this involves dispatching on the type of `x[i]`; in object-oriented languages,
+it might involve dispatching on the type of `f`), push the argument `x[i]`
+etcetera to `f` via a register and/or a [call stack](https://en.wikipedia.org/wiki/Call_stack),
+jump to the machine instructions to execute them, jump back to
+the caller `naivebroadcast`, and extract the return value.
+That is, calling a function argument `f` involves some overhead beyond
+the cost of the computations inside `f`.
+
+If `f(x)` is expensive enough, then the overhead of the function call may be negligible,
```
nalimilan: It also prevents SIMD vectorization, which is a big deal for standard arithmetic operations.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,449 @@
...
+* You have to decide *in advance* whether a given function `f(x::Number)`
+  will also be applied elementwise to arrays, and define a corresponding elementwise method.
+
+For library functions like `sqrt`, this means that the library authors
+have to guess at which functions should have vectorized methods, and users
+have to guess at what vaguely defined subset of library functions work
+for vectors.
+
+One possible solution is to vectorize *every function automatically*.   The
+language [Chapel](https://en.wikipedia.org/wiki/Chapel_%28programming_language%29)
```
stevengj: Not the same thing, I think.  In Matlab, scalars are 1-element vectors too, but this is not the same thing as what Chapel does either.   In Julia, you can declare a function as `f(x) = ...` and it will accept any argument type, but it won't necessarily *work* for any argument type.

The problem is that the same code doesn't generally work for scalars as for vectors.   e.g. in the example in the blog post, `f(x) = 3x^2 + 5x + 2`doesn't work for vectors because `x^2` isn't defined for vectors.   In R, if you have a function like `if (x < 0) { 2*x } else { 3*x }` then it will do *something* for vectors, but probably not what you want (it will only look at the first element of `x < 0`).

Chapel, in contrast, effectively defines a generic function `(f::Function)(args::AbstractArray...) = map(f, args...)` that is called as a fallback, so if you just define `f(x::Number) = x < 0 ? 2*x : 3*x` it will do the elementwise thing for `f(::Array)`.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,449 @@
...
+dot calls `(+).(x,y)`.   (Unlike Chapel's implicit vectorization, Julia's
+`f.(x...)` syntax corresponds to `broadcast(f, x...)` rather than `map`,
+which allows you to combine arrays of different shapes/dimensions.)
+From the standpoint of the programmer, this adds a certain amount of
+clarity because it indicates explicitly when an elementwise operation
+is occuring.  From the standpoint of the compiler, dot-call syntax
+enables the *syntactic loop fusion* optimization described in more detail
+below, which we think is an overwhelming advantage of this style.
+
+### Why vectorized code is fast
```
stevengj: Yes, will do.

blog/_posts/2016-12-xx-moredots.md
stevengj: Does broadcast do `@simd`, though?

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,449 @@
...
+    x = X[i]
+    X[i] = f(2x^2 + 6x^3 - sqrt(x))
+end
+```
+
+(Of course, like all Julia code, to get good performance both of these snippets should be executed inside some function, not in global scope.)
+
+In this blog post, we delve into some of the details of this new development, in order to answer questions that often arise when this feature is presented:
+
+* What is the overhead of traditional "vectorized" code?  Isn't vectorized code supposed to be fast already?
```
stevengj: will do

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,449 @@
...
+e.g. `f(x) = 3x.^2 + 5x + 2` (using the elementwise operator `.^` because
+`array^scalar` is not defined).   (If we cared a lot about efficiency, we
+might instead define a special method `f(X::AbstractArray) = map(f, X)`
+or use the `@vectorize_1arg f Number` macro defined in Julia 0.4.)
+
+### Which functions are vectorized?
+
+As an aside, this example illustrates an annoyance with the vectorized style:
+
+* You have to decide *in advance* whether a given function `f(x::Number)`
```
stevengj: ok

blog/_posts/2016-12-xx-moredots.md
stevengj: will clarify

blog/_posts/2016-12-xx-moredots.md
stevengj: Will add a mention.

stevengj: Added a comment.

stevengj:
@ViralBShah, I'll try to put together a notebook.

blog/_posts/2016-12-xx-moredots.md
johnmyleswhite: This seems good to me now. I might rephrase it as:

"In particular, loading data from main computer memory (RAM) is much slower than performing arithmetic operations on data stored in registers, so recently used data is [...]"

blog/_posts/2016-12-xx-moredots.md
nalimilan: Yeah, I'm not arguing it's the same thing, nor that it's preferable. But it's true that in R in many cases people write vectorized methods without even thinking about it -- until that breaks for non-trivial cases.

blog/_posts/2016-12-xx-moredots.md
stevengj: Thanks, I've adopted a version of this suggestion.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,456 @@
...
+
+As an aside, this example illustrates an annoyance with the vectorized style:
+you have to *decide in advance* whether a given function `f(x)`
+will also be applied elementwise to arrays, and either
+write it specially or define a corresponding elementwise method.
+
+(Even if a function *accepts* an array argument `x` — our function `f`
+accepts any `x` type, and in Matlab or R there is no distinction between
+a scalar and a 1-element array — that doesn't mean it will *work* elementwise
+for an array unless you write the function with that in mind.)
```
Sacha0: The separation between this sentence's pre- and post- emdash-parenthetical-phrase parts caused my wetware parser to throw and retry a couple times. Perhaps
```julia
Even if a function *accepts* an array argument `x`, that doesn't mean it will *work* elementwise for an array (unless you write the function with that in mind). For example, our function `f`'s original definition (`f(x) = 3x^2 + 5x + 2`) accepts any `x` type, but does not work for arrays.
```
would be easier to follow reading linearly? (With the `Matlab`/`R` mention integrated somehow of course.) Best!

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,456 @@
...
+
+Instead, starting in Julia 0.5, *any* function `f(x)` can be applied elementwise
+to an array `X` with the ["dot call" syntax `f.(X)`](http://docs.julialang.org/en/stable/manual/functions/#dot-syntax-for-vectorizing-functions).
+Thus, the *caller* decides which functions to vectorize.  In Julia 0.6,
+"traditionally" vectorized library functions like `sqrt(X)` are [deprecated](https://github.com/JuliaLang/julia/pull/17302) in
+favor of `sqrt.(X)`, and dot operators
+like `x .+ y` are [now equivalent](https://github.com/JuliaLang/julia/pull/17623) to
+dot calls `(+).(x,y)`.   (Unlike Chapel's implicit vectorization, Julia's
+`f.(x...)` syntax corresponds to `broadcast(f, x...)` rather than `map`,
+which allows you to combine arrays of different shapes/dimensions.)
```
Sacha0: Whether the "which" applies to `broadcast(f, x...)` or `map` seems ambiguous. Perhaps "which allows" -> "allowing" or "rather than `map`" -> ("rather than `map`")?

Sacha0: At this point [`@pabloferz`'s comment](https://github.com/JuliaLang/julialang.github.com/pull/480#issuecomment-269572547) occurred to me as well. Perhaps an example of `broadcast`'s shape/dimension flexibility would do well here?

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,456 @@
...
+potentially build the compiler to recognize that it can fuse
+`*`, `+`, `.^`, and similar operations for the built-in `Array` type,
+but only for elements that are one of a few built-in numeric types (`Float64`, `Int`, etc.).
+These operations are common enough that this may still be a worthwhile optimization
+for Julia to implement at some point!
+
+However, even though Julia will certainly implement additional compiler
+optimizations as time passes, one of the key principles of Julia's design
+is to "build in" as little as possible into the core language, implementing
+as much of possible of Julia *in Julia* itself [(Bezanson, 2015)](https://github.com/JeffBezanson/phdthesis/blob/master/main.pdf).
```
Sacha0: Perhaps "as much of possible" -> "as much as possible"?

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,456 @@
...
+indicates, by adding dots, which function calls and operators are
+intended to be applied elementwise (specifically, as `broadcast` calls).
+The compiler notices these dots at *parse time* (or technically
+at "lowering" time, but in any case long before it knows
+the types of the variables etc.), and transforms them into calls to
+`broadcast`.  Moreover, it guarantees that *nested* "dot calls" will
+*always* be fused into a single broadcast call, i.e. a single loop.
+
+Put another way, `f.(g.(x .+ 1))` is treated by Julia as simply
+[syntactic sugar](https://en.wikipedia.org/wiki/Syntactic_sugar) for
```
Sacha0: Perhaps "as simply syntactic sugar" -> "simple as syntactic sugar"?

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,456 @@
...
+### Other partway solutions
+
+For the sake of completeness, we should mention
+some other possibilities that would partly
+address the problems of vectorization.  For example, functions could
+be specially [annotated to declare that they are pure](https://github.com/JuliaLang/julia/issues/414),
+one could specially annotate container types with
+array-like semantics, etcetera, to help the compiler recognize the
+possibility of fusion.   But again, this imposes a lot of requirements
+on the library authors, and as above it requires them to identify
```
Sacha0: No need for the article preceding "library authors"?

blog/_posts/2016-12-xx-moredots.md
stevengj: typo, thanks

blog/_posts/2016-12-xx-moredots.md
stevengj: as merely?

blog/_posts/2016-12-xx-moredots.md
stevengj: will break into two sentences

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,456 @@
...
+author: <a href="http://math.mit.edu/~stevenj">Steven G. Johnson</a>
+---
+
+After a [lengthy design process](https://github.com/JuliaLang/julia/issues/8450) and [preliminary foundations in Julia 0.5](http://julialang.org/blog/2016/10/julia-0.5-highlights#vectorized-function-calls), Julia 0.6 includes new facilities for writing code in the "vectorized"
+style (familiar from Matlab, Numpy, R, etcetera) while avoiding the
+overhead that this style of programming usually imposes: multiple
+vectorized operations can now be "fused" into a single loop, without
+allocating any extraneous temporary arrays.
+
+This is best illustrated with an example.  Suppose we have
```
Sacha0: This post is fantastic. The more who read it through, the better. In the interest of hooking readers as soundly as possible from the get-go, augmenting this introductory example with a simple, stark performance "wow!" would be great. (I wager that an explicit runtime reduction illustration will better capture most readers' attention than the existing description of fusion into a single loop --- to have that description hit home as a "wow!", the reader need already appreciate most of the content of this post.) Thoughts? Best!

blog/_posts/2016-12-xx-moredots.md
Sacha0: Sounds good!

blog/_posts/2016-12-xx-moredots.md
stevengj: Sure, I'll mention the factor of 10 in performance and memory up-front

stevengj:
Added a notebook, with placeholders in the blog post to link to the notebook where appropriate, with comparisons between devectorized, traditional vectorized, traditional vectorized + preallocation, new-style fused vectorized, and C code.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,468 @@
...
+```jl
+X = f(2 * X.^2 + 6 * X.^3 - sqrt(X))
+```
+
+Of course, this assumes that the functions `sqrt` and `f` are "vectorized,"
+i.e. that they accept vector arguments `X` and compute the
+function elementwise.  This is true of `sqrt` in Julia 0.4, but it
+means that we have to rewrite our function `f` from above in a vectorized style, as
+e.g. `f(x) = 3x.^2 + 5x + 2` (changing `f` to use the elementwise operator `.^` because
+`array^scalar` is not defined).   (If we cared a lot about efficiency, we
```
tkelman: maybe say vector^scalar, since matrix^scalar is defined

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,468 @@
...
+### Why vectorized code is fast
+
+In many dynamic languages for technical computing, vectorization is
+seen as a key (often *the* key) performance optimization.   It allows
+your code to take advantage of highly optimized (perhaps even parallelized)
+library routines for basic operations like `scalar*array` or `sqrt(array)`.
+Those functions, in turn, are usually implemented in a low-level language
+like C or Fortran.   Writing your own "devectorized" loops, in contrast,
+is too slow, unless you are willing to drop down to a low-level language
+yourself, because the semantics of those languages make it hard to
```
tkelman: "those languages" isn't obviously about the "dynamic languages for..." on first reading

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,468 @@
...
+
+One approach that may occur to you, and which has been implemented in a
+variety of languages (e.g. [Kennedy & McKinley, 1993](http://dl.acm.org/citation.cfm?id=665526);
+[Veldhuizen, 1995](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.43.248);
+[Lewis et al., 1998](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.46.6627);
+[Chakravarty & Keller, 2001](http://dl.acm.org/citation.cfm?id=507661);
+[Manjikian & Abdelrahman, 2002](http://ieeexplore.ieee.org.libproxy.mit.edu/document/577265/);
+[Sarkar, 2010](http://ieeexplore.ieee.org/document/5389392/);
+[Prasad et al., 2011](http://dl.acm.org/citation.cfm?id=1993517);
+[Wu et al., 2012](http://dl.acm.org/citation.cfm?id=2457490)), is to only
```
tkelman: Maybe worth also referencing ParallelAccelerator and Numba here as easily accessible implementations? I believe they both do some form of this.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,468 @@
...
+
+In typical performance benchmarks (LINK NOTEBOOK), therefore, the traditional
+vectorized code `X = f(2 * X.^2 + 6 * X.^3 - sqrt(X))` turns out to be **about
+10× slower** than the devectorized or fused-vectorized versions of the same code
+at the beginning of this article for `X = zeros(10^6)`.   Even if we
+pre-allocate all of the temporary arrays (completely eliminating the allocation
+cost),  our benchmarks show that performing a separate loop for each operation
+still is about 4–5× slower for a million-element `X`. This is not unique to
+Julia!  **Unfused vectorized code is suboptimal in any language** unless the
+language's compiler can automatically fuse all of these loops (even ones that
```
tkelman: if the compiler automatically fuses it, does it still count as unfused code?

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,468 @@
...
+The second and third steps pose an *enormous challenge*: looking at an arbitrary
+function and "understanding" it at this level turns out to be a very hard
+problem for a computer.  If fusion is viewed as a compiler *optimization*, then the
+compiler is only free to fuse if it can *prove* that fusion *won't change the
+results*, which requires the detection of purity and other data-dependency
+analyses.
+
+In contrast, when the Julia compiler sees an expression like `2 .* X .+ Y`,
+it knows just from the *syntax* (the "spelling") that these are elementwise
+operations, and it *guarantees* that it will *always* fuse into a single
```
tkelman: should maybe clarify what the "it" is that's responsible for the guarantee

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,468 @@
...
+languages as discussed above; would you then be able to
+implement the same fusing semantics efficiently?
+
+There is a catch: `2 .* x .+ x .^ 2` is sugar for
+`broadcast(x -> 2*x + x^2, x)` in Julia, but for this to be
+fast we need the [higher-order function](https://en.wikipedia.org/wiki/Higher-order_function)
+`broadcast` to be very fast as well.  First, this
+requires that arbitrary user-defined scalar (non-vectorized!) functions like
+`x -> 2*x + x^2` be compiled to fast code, which isn't true in
+most other dynamically typed high-level languages like Python, but is usually true
```
tkelman: too much of a rabbit hole, but it would by in pypy, wouldn't it?

blog/_posts/2016-12-xx-moredots.md
stevengj: Yes, I want to move the mention of DSLs like Theano and PyOP2 to here, and ParallelAccelerator is definitely something I should mention.  I didn't realize that Numba does loop fusion, but that seems to have been implemented in Numba in 2015.

stevengj:
@pabloferz, I added a section at the end reviewing the generality of `broadcast`: combining containers of different shapes, performing non-numeric computations, and even fusing operations that aren't really on "containers" at all (JuliaApproximation/ApproxFun.jl#356).

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -458,3 +464,135 @@ higher-order functions are a key ingredient of Julia that allows
...
+`6 .* X.^3` that combine an array (`X`) with scalars (`6` and `3`).
+Conceptually, in `X.^3` the scalar `3` is "expanded" (or "broadcasted")
+to match the size of `X`, as if it became an array `[3,3,3,...]`,
+before performing `^` elementwise.  In practice of course, no array
+of `3`s is ever explicitly constructed.
+
+More generally, if you combine two arrays of different dimensions or shapes,
+any "singleton" (length 1) or missing dimension of one array is "broadcasted"
+across that dimension of the other array.   For example, `A .+ [1,2,3]`
+adds `[1,2,3]` to *each column* of an *n*×3 matrix `A`.   Another typical
```
tkelman: 3-by-n?

stevengj: fixed

blog/_posts/2016-12-xx-moredots.md
stevengj: (Fixed.)

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,600 @@
...
+layout: post
+title:  More Dots: Syntactic Loop Fusion in Julia
+author: <a href="http://math.mit.edu/~stevenj">Steven G. Johnson</a>
+---
+
+After a [lengthy design process](https://github.com/JuliaLang/julia/issues/8450) and [preliminary foundations in Julia 0.5](http://julialang.org/blog/2016/10/julia-0.5-highlights#vectorized-function-calls), Julia 0.6 includes new facilities for writing code in the "vectorized"
+style (familiar from Matlab, Numpy, R, etcetera) while avoiding the
+overhead that this style of programming usually imposes: multiple
+vectorized operations can now be "fused" into a single loop, without
+allocating any extraneous temporary arrays.
```
pabloferz: "...allocating extraneous temporary arrays." or "allocating any extraneous temporary array."?

stevengj: "any arrays" seems more accurate, and is grammatical.   Typically, more than one temporary array would be allocated without fusion.

blog/_posts/2016-12-xx-moredots.md
```diff
@@ -0,0 +1,600 @@
...
+the fit. (Conceptually, this replaces *elementwise* operations on containers
+with *pointwise* operations on functions.) In contrast, ApproxFun also allows
+you to compute the same result using `exp(f^2 + f^3)`, but in this case it will go
+through the fitting process *four times* (constructing four `Fun` objects), once
+for each operation like `f^2`, and is more than an order of magnitude slower
+due to this lack of fusion.
+
+### `broadcast` vs. `map`
+
+It is instructive to compare `broadcast` with `map`, since `map` *also*
```
pabloferz: "Finally, it is instructive..." ?

blog/_posts/2016-12-xx-moredots.md
pabloferz: All right :+1:

stevengj:
Hmm, I ran the notebook on my desktop machine, and got a bit more overhead for the dot-vectorized version (times are normalized to the devectorized version):
```
       "n"    "devec"    "vec"     "vec_pre"   "newvec"
      1.0    1.0       37.5875   14.6582      1.60781  
      5.0    1.0       20.198     8.08684     1.62422  
     10.0    1.0       11.4514    4.86414     1.43233  
     20.0    1.0        6.62008   2.86636     1.32749  
     50.0    1.0        3.92799   1.85227     1.25438  
    100.0    1.0        2.4905    1.41182     1.21229  
   1000.0    1.0        2.27151   1.26594     1.20959  
   5000.0    1.0        2.00991   1.85677     1.20218  
  10000.0    1.0        2.16502   1.89519     1.19052  
 100000.0    1.0        3.42414   2.91594     1.19419  
      1.0e6  1.0        7.92444   5.3167      1.18219  
      1.0e7  1.0       16.9037    5.28106     1.18397  
```
@pabloferz, does 20% overhead sound right to you?  Can we do better?

ViralBShah:
Merge?

tkelman:
date is no longer accurate

pabloferz:
@stevengj That's certainly a lot of overhead, but it doesn't come from the broadcast machinery itself. It has to do with the lowering of the dot-calls. It turns out that the compiler can't optimize too well the composed anonymous functions generated at lowering. A way to see this is comparing the performance of `devec!` and `newvec!` for a long and complex function `g`. E.g.,

```julia
f(x) = 3x.^2 + 5x + 2
g(x) = f(2x^2 + 6x^3 - sqrt(x))

function devec!(X)
    for i in eachindex(X)
        x = X[i]
        X[i] = g(x)
    end
    return X
end

newvec!(X) = X .= g.(X)
```

Both should give you similar performance (`newvec!` might be even slightly faster).

ViralBShah:
@stevengj I assume you will change the date and merge. Would love to see this go out.

stevengj:
Merged.  Will it show up automatically?

ViralBShah:
It should show up automatically, but is not. Also, something funny is going on in the blog display for the first two entries:

http://julialang.org/blog/

@shashi Thoughts?

ViralBShah:
I wonder if the colon in the title is throwing it off.

blog/_posts/2017-01-21-moredots.md
```diff
@@ -0,0 +1,600 @@
...
+  compiler optimization that may or may not occur for carefully written code.  They also allow the *caller* to "vectorize" *any* function, rather than relying on the function author.
+
+3. Other languages have implemented loop fusion for vectorized operations,
+  but typically for only a small set of types and operations/functions that
+  are known to the compiler.  Julia's ability to do it generically, even
+  for *user-defined* array types and functions/operators, is unusual
+  and relies in part on the syntax choices above and on its ability to efficiently
+  compile higher-order functions.
+
+Finally, we'll review that why, since these dots actually correspond to
```
Sacha0: Should "review that why" be "review why"?

stevengj: fixed

stevengj:
Do we announce blog posts on discourse or something?

tkelman:
may as well! and people can post it to hacker news, reddit etc

stevengj:
I posted an announcement on discourse; I'll leave it to others to deal with hacker news etc.

shashi:
@ViralBShah #492 fixed it. If you have a colon in your title YAML parsing on github pages gets messed up - use quotes or switch to `-`.

Also, I love this post!

PR #487, opened 2017-01-16
Modify DiffEq projects: natural syntax parsing

ChrisRackauckas:
One last modification to the JuliaDiffEq projects. I removed the JuliaDiffEq plotting project because overtime other contributors keep chipping away at it so I don't think there's a summer left there. However, this new project is a very good opportunity. We will be launching DifferentialEquations.jl online soon:

http://app.juliadiffeq.org/

and realized that we need to expand the functionality of ParameterizedFunctions.jl to be able to parse delay differential equations and differential-algebraic equations. So the start of the project is simple to describe: expand the parse-able syntax of the ParameterizedFunctions.jl macros to allow for these types of equations, but an advanced student can go a lot of places from here, including:

- Add DAEs and DDEs to DifferentialEquations.jl Online
- Allow the macros to perform automatic index reduction on DAEs
- Modify the symbolic calculations within the macro to define extensions to the Jacobians etc. for DAE solvers
- Provide the ability to reduce some DDEs to ODEs
- Check if the differential equation has a solution and, if it does, return that analytical solution
- Expand DiffEqProblemLibrary to contain many examples, both for testing and as a tutorial

Given that symbolic calculations are already performed in the macro using SymEngine.jl, there's essentially a tutorial for how to get started on all of these already within the source code. Thus the resulting project can go in more web, mathematical, computer science (parsing), or even scientific directions depending on the student.

(Note: do not share the DifferentialEquations.jl Online weblink yet as we still have some work to do before launching. Thus for now this is labelled WIP)

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -177,6 +177,41 @@ This project proposal is to implement a native Julia framework for distributed e
 
 # Theme: Numerical Methods for Differential Equations
 
+## DiffEq-specific differentiation library to ease development of fast native stiff solvers
+
+Methods for solving stiff differential equations and differential algebraic equations
+(DAEs) are composed of similar calculus objects: gradients, Jacobians, etc. However,
+not only are these objects common, but there are also very common use patterns.
+For example, the quantity `(M - gamma*J)^(-1)`, (where `M` is a mass matrix, `J`
+is the Jacobian, and `gamma` is a constant), is found in Rosenbrock methods,
+Newton solvers for implicit Runge-Kutta methods, and update equations for
+implicit multistep methods. However, the current setup requires that every solver
+implements these functions on their own, leaving the chosen options and optimizations
+as repeated work throughout the ecosystem.
+
+The goal would be to create a library which exposes many different options
+for computing these quantities to allow for easier development of native solvers.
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
DiffEqOnline has been released so this can be merged now.

MikeInnes:
This has a conflict because I reorganised the ideas page. Given that JuliaDiff is its own org with a large number of projects, I figured it might be better for you to host your own page somewhere and we can just link to it from here.

That's not a unilateral decision though, if you'd rather host the projects here I'll happily add them back in with this patch.

ChrisRackauckas:
I think it might be better to have it all in one spot. The ToC at the top does a good job at separating. But yeah, this setup is not scaling well in terms of Git: changes to the ideas page does always cause a merge conflict. I think we should consider a slightly new design for the page, where each theme is a page? (And add some more machine learning projects, I'll find someone for that)

MikeInnes:
Did you see the newest version? I'm pretty sure I just did what you're thinking of :) More ML would be great.

Will add the diff page back in.

ChrisRackauckas:
Oh ha, nice work!

MikeInnes:
Ok, actually I have to sign off for the day – I can do it tomorrow, or I'll merge this PR if you fix it before then.

ChrisRackauckas:
I can do it right now.

PR #489, opened 2017-01-20
Edit on Github link

MikeInnes:
Was just going to edit the gsoc page quickly and fell down a rabit hole. Anyway, this adds an "edit on github" link at the bottom of pages to make that a bit quicker / more accessible.

Also fixes a couple of urls to use relative references. This makes it easier to test a fork like [this](https://mikeinnes.github.io/julialang.github.com/); all the links and CSS don't break.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -23,7 +23,7 @@ navigation:
     title: teaching
   - url: "/publications/"
     title: publications
-  - url: "/soc/"
+  - url: "/soc/ideas-page.html"
     title: GSoC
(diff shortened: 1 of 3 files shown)


MikeInnes:
Pushed this by accident. Should be a pretty uncontroversial change though regardless.

MikeInnes:
Ah, but not the actual link bit.

tkelman:
please be more careful about what branches you're pushing to. I think you completely deleted the differential equations projects.

PR #490, opened 2017-01-20
Add DiffEq

ChrisRackauckas:
This has it as a diffeq page. Or should it be a scientific applications page which includes optimization and machine learning projects?


MikeInnes:
:+1: this organisation is fine for now; we can tweak it as and when those other areas grow

PR #491, opened 2017-01-21
Fix link

lpuettmann:
The link in the [blog post](http://julialang.org/blog/2017/01/2017-01-21-moredots) is broken and I suspect this is what was intended.


pabloferz:
Yeah, that's the one. Thank you!

PR #492, opened 2017-01-21
Remove colon from More Dots blog post title

johnmyleswhite:
This attempts to fix the rendering problems by replacing a colon with a dash. Judging from previous posts (e.g. http://julialang.org/blog/2015/10/datastreams), we should also have been to get the colon to render by placing the title in quotes.

### DIFF
--- a/blog/_posts/2017-01-21-moredots.md
+++ b/blog/_posts/2017-01-21-moredots.md
@@ -1,6 +1,6 @@
 ---
 layout: post
-title:  More Dots: Syntactic Loop Fusion in Julia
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
This issue seems to confirm my understanding of this issue: https://github.com/jekyll/jekyll/issues/549

I think we should merge this as soon as someone else verifies I didn't make things worse.

johnmyleswhite:
Thanks, @MikeInnes!

MikeInnes:
Seemed to work, thank you!

PR #498, opened 2017-02-06
Add post on custom array indices

timholy:
This isn't ready yet, but I'm opening as a PR now to make it easier for anyone to comment or ask questions at specific points in the text.


blog/_posts/2017-02-05-offset-arrays.md
```diff
@@ -0,0 +1,489 @@
...
+julia> A[3,4]
+18
+```
+
+So the indices in `B3` match those of `A`. Indeed, `B3` doesn't even
+have an element "named" `(1,1)`:
+
+```julia
+julia> B3[1,1]
+ERROR: BoundsError: attempt to access OffsetArrays.OffsetArray{Int64,2,Array{Int64,2}} with indices 2:4×1:4 at index [1,1]
```
KristofferC: Perhaps use v0.6 stacktraces in the final version of the post?

timholy: Good idea. Esp. since 0.5.0 (rather than release-0.5) has some significant bugs when it comes to `indices` support.

ViralBShah:
Bump. In case this is ready to merge.

blog/_posts/2017-02-05-offset-arrays.md
```diff
@@ -0,0 +1,489 @@
...
+
+If we display these three images, we get the following:
+
+![mandrill](https://github.com/JuliaLang/julialang.github.com/blob/master/blog/_posts/offsetarrays_figures/mandrill.png?raw=true)
+![mandrill](https://github.com/JuliaLang/julialang.github.com/blob/master/blog/_posts/offsetarrays_figures/mandrill_eye.png?raw=true)
+![mandrill](https://github.com/JuliaLang/julialang.github.com/blob/master/blog/_posts/offsetarrays_figures/mandrill_nose.png?raw=true)
+
+In summary, it's very straightforward to pick a particular region of
+an array and extract it for further analysis.  Like the Talking Heads
+song "Road to Nowhere," we know where we're going...
```
rsrock: Question for you, Tim (related to what follows).  Let's say I wanted to extract a region from an image into a buffer.  I'm going to do a lot of expensive calculations on that region, so it makes sense to copy that data into a buffer instead of using a view into the original image.  Let's also say that the buffer will be a fixed size, but the target location will change once I'm done processing one object.  For example, I grab the left eye, calculate, then grab the right eye and calculate.  

Can I mutate the ranges used to define the offset array (keeping the array size fixed) so I don't have to fill up the garbage collector with a whole bunch of buffers?

timholy: You can definitely re-use the buffer. You have to create a new OffsetArray wrapper. Something like this:
```julia
buf = Matrix{T}(30, 30)
chunk = OffsetArray(buf, 201:230, 201:230)
R = CartesianRange(indices(chunk))
copy!(chunk, R, img, R)
# do something useful
chunk = OffsetArray(buf, 201:230, 401:430)
R = CartesianRange(indices(chunk))
copy!(chunk, R, img, R)
# do something useful
```
Does that answer your question?

You may know this already, but for reasons of safety we don't allow turning off bounds-checking on OffsetArrays; you'll want the `@unsafe` macro if you need high performance. By Julia 1.0 I think we should turn on the ability to remove bounds-checks. This blog post will hopefully get more people testing non-1 arrays so more of our code is ready.

rsrock: Perfect, thanks for the tip.

timholy:
OK, I reworked the introductory example to exploit the work that @Evizero and I have been doing in ImageTransformations. I think that makes a more compelling first example.

Is there a good way to preview the result? The image links are specific for the paths here, and there's math formatting. If not, let's just merge it and I can fix the issues before I advertise the new post.

blog/_posts/2017-02-05-offset-arrays.md
```diff
@@ -0,0 +1,620 @@
...
+
+The fact that the output preserves the coordinates makes it trivial to
+compare the images:
+
+```julia
+julia> cv = colorview(RGB, paddedviews(0, img, img_rotated, img)...)
+```
+
+`paddedviews(0, arrays...)` pads input arrays with 0, as needed, to
+given them all the same indices, and `colorview(RGB, r, g, b)` inserts
```
rfourquet: given -> give (?)

timholy:
If anyone besides me ends up merging this, my intention is to squash into a single commit.

ViralBShah:
Best if you merge it whenever you are ready.

timholy:
That makes sense; that way I can be on the lookout for formatting errors and fix them quickly.

I might wait until I tag https://github.com/JuliaIO/ImageMagick.jl/pull/85 (hopefully in a day or two).

KristofferC:
The way I made my images, they are at least showed in the markdown preview: https://github.com/JuliaLang/julialang.github.com/blob/kc/repl0.6/blog/_posts/2017-02-22-repl-0.6-highlights.md

ararslan:
Woot! Great to see this merged. Excellent work here, Tim!

timholy:
Well, the images are all borked.

timholy:
OK, fixed now (see #556). I'll wait a few hours before I advertise this on Discourse, in case anyone sees problems with the "published" product that didn't show up during review.

PR #499, opened 2017-02-06
Update link to user discussion forum.

christi3k:
Fixes link to user support in Github section by updating to Usage category on Julia Discourse.


ararslan:
Great, thank you!

PR #502, opened 2017-02-17
Change to link to SIAM article

simonbyrne:
I'm assuming these are they same article? You may also want to add the SIAM link from the arxiv page.


simonbyrne:
I realised the abstracts are different as well: should these be updated?

ViralBShah:
Yes, they are the same paper. During final publication, the abstract probably changed slightly. Cc @alanedelman

ViralBShah:
Merging since this is a strict improvement.

PR #504, opened 2017-02-20
Update QML description

barche:
The QML bindings description was still from last year. In the mean time, I feel my QML.jl package is mature enough to be proposed as a topic. I'd be willing to mentor if needed.


MikeInnes:
That would be great, we'd certainly love to have you as a mentor for a QML project. Thanks for the project update!

PR #507, opened 2017-02-22
Add recommended skills and expected results

ChrisRackauckas:
I added a bunch  of recommended skills and expected results for the numerics page. In general, those should be about all that's needed for the projects. I tried to really emphasize the difference of "for this, you just need to know calculus/linear algebra well, and be familiar enough with this other thing to pick it up as you go long" vs "you might need to know some (insert non-calculus/linear algebra) to do this well"


MikeInnes:
Perfect, thanks for these improvements.

PR #514, opened 2017-02-27
gsoc section: Data Structure Algorithm Improvements

vtjnash:


ChrisRackauckas:
Every project needs recommended skills, expected results, and mentors who would be willing.

shashi:
@vtjnash edited to add skills and mentors. Hit merge if it's OK

vtjnash:
I'm probably not a good mentor for this. If nobody volunteers their name here, it's probably best to leave this in PR state.

MikeInnes:
This project is pretty general; we likely don't need the world's foremost data structure expert to mentor it. Let's leave it without a mentor and we can try to match a good student with someone who's free.

PR #515, opened 2017-02-27
gsoc section: move BigNum section from Compiler -> Numerics

vtjnash:


MikeInnes:
The projects are meant to be organised by skillsets more than by use cases. By that standard I'd say the bignums project is a lot closer to "compiler" than to "numerics".

Admittedly it's still something of a misfit given that most of the other projects here are on Julia's internals. Maybe general would be a better place?

vtjnash:
Possibly. I left most of the current text in, but was attempting to more transition it towards "write an optimized bigint library in pure Julia" rather than "create a second, mutable BigInt type"

ChrisRackauckas:
I don't know about categorizing by skillsets and not topics. If that's the case, some other things would for sure need to be moved around. But I think it's more sensible to find a project of interest by looking into a category/field which you find interesting, and browsing the topics in there.

MikeInnes:
There are arguments for both sides, but I suspect that students are going to come to the numerics page because of an interest in, say, numerical stability or mathematical optimisation; a project involving wrapping data structures and implementing object pooling isn't then likely to be a great fit for that student.

That said, I hadn't noticed @vtjnash's tweaks and a reimplementation seems like it'd involve more numerical issues, so fair enough.

PR #520, opened 2017-03-03
Add JuliaPro as a way to download Julia.

ViralBShah:
We've been working on JuliaPro for a while, and it is now quite robust and easy to use and install. Adding it here as an option for others to have an alternative way.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -12,9 +12,9 @@ We provide several ways for you to run Julia:
 * [JuliaPro](http://juliacomputing.com/products/juliapro.html) by [Julia Computing](http://juliacomputing.com) is a batteries included distribution of Julia. It includes the [Juno IDE](http://junolab.org), the [Gallium debugger](https://github.com/Keno/Gallium.jl), and a number of packages for plotting, optimization, machine learning, databases and much more (requires registration).
 
 Plotting capabilities are provided by external packages such as
-[PyPlot.jl](https://github.com/stevengj/PyPlot.jl) and [Gadfly.jl](http://gadflyjl.org).
+[PyPlot.jl](https://github.com/JuliaPy/PyPlot.jl) and [Gadfly.jl](http://gadflyjl.org).
 A package which integrates most of Julia's plotting backends into one convenient and
-well-documented API is [Plots.jl](https://github.com/tbreloff/Plots.jl). Look at the
+well-documented API is [Plots.jl](https://github.com/JuliaPlots/Plots.jl). Look at the
 [plotting instructions](plotting.html) to install a plotting package. If you are using
 JuliaBox, all of these plotting packages are pre-installed.
 
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -8,14 +8,14 @@ title:  Julia Downloads
 We provide several ways for you to run Julia:
 
 * In the terminal using the built-in Julia command line.
-* The [Juno](http://www.junolab.org) integrated development environment (IDE).
 * In the browser on [JuliaBox.com](https://www.juliabox.com) with Jupyter notebooks. No installation is required -- just point your browser there, login and start computing.
+* The [JuliaPro distribution](http://juliacomputing.com/products/juliapro.html) by Julia Computing provides a single installer for Julia, the [Juno IDE](http://junolab.org), the Gallium debugger, and a number of packages (requires registration).
 
 Plotting capabilities are provided by external packages such as
-[PyPlot.jl](https://github.com/stevengj/PyPlot.jl) and [Gadfly.jl](http://gadflyjl.org). 
-A package which integrates most of Julia's plotting backends into one convenient and 
-well-documented API is [Plots.jl](https://github.com/tbreloff/Plots.jl). Look at the 
-[plotting instructions](plotting.html) to install a plotting package. If you are using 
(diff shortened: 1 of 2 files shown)


ViralBShah:
We used to have an old Juno based Julia distribution that is no longer maintained, and I have removed traces of that. JuliaPro achieves the same goal now.

downloads/index.md
```diff
@@ -8,14 +8,14 @@ title:  Julia Downloads
 We provide several ways for you to run Julia:
 
 * In the terminal using the built-in Julia command line.
-* The [Juno](http://www.junolab.org) integrated development environment (IDE).
 * In the browser on [JuliaBox.com](https://www.juliabox.com) with Jupyter notebooks. No installation is required -- just point your browser there, login and start computing.
+* The [JuliaPro distribution](http://juliacomputing.com/products/juliapro.html) by Julia Computing provides a single installer for Julia, the [Juno IDE](http://junolab.org), the Gallium debugger, and a number of packages (requires registration).
```
ararslan: It might be nice to make Julia Computing a link to the Julia Computing website and similarly for Gallium to https://github.com/Keno/Gallium.jl.

downloads/index.md
```diff
@@ -8,14 +8,14 @@ title:  Julia Downloads
...
 
 Plotting capabilities are provided by external packages such as
-[PyPlot.jl](https://github.com/stevengj/PyPlot.jl) and [Gadfly.jl](http://gadflyjl.org). 
-A package which integrates most of Julia's plotting backends into one convenient and 
-well-documented API is [Plots.jl](https://github.com/tbreloff/Plots.jl). Look at the 
-[plotting instructions](plotting.html) to install a plotting package. If you are using 
+[PyPlot.jl](https://github.com/stevengj/PyPlot.jl) and [Gadfly.jl](http://gadflyjl.org).
```
ararslan: While you're in here it might be nice to update the URL for PyPlot, as it's now in the JuliaPy organization.

ChrisRackauckas:
I kind of liked that Juno was mentioned. But I think what that's pointing to is that we might want to have a separate section that points users to some standard tooling, like IDEs, plotting, etc. Right now it's kind of mixed in there, but should probably be "right after":

- Here's some easy ways to get Julia.
- Here's the way to get "basic Julia".
- Here's some standard tooling.

Otherwise I don't think there is a mention of the fact that there is an IDE (Juno) for Julia other than in that one line about JuliaPro. I think it might come off as though JuliaPro and Juno are somehow linked/closer, and maybe some might think you need JuliaPro to use Juno.

downloads/index.md
```diff
@@ -8,14 +8,14 @@ title:  Julia Downloads
...
+A package which integrates most of Julia's plotting backends into one convenient and
+well-documented API is [Plots.jl](https://github.com/tbreloff/Plots.jl). Look at the
```
ararslan: Similarly, Plots now lives in the JuliaPlots organization

ViralBShah:
I think the point is valid - that there is some information about what the user may want to do after downloading Julia, and it is all mixed up here. Open to ideas. Perhaps we should have a section on what to do after downloading, where we can be elaborate in pointers to various things.

ViralBShah:
Actually, I don't think this plotting stuff should be here at all - and it should actually be in the post install section.

ViralBShah:
I updated the URLs for the plotting packages for now. The post install thing should be a separate PR anyways. Any takers?

downloads/index.md
ViralBShah: Done.

downloads/index.md
ViralBShah: Done.

downloads/index.md
ViralBShah: Done.

ViralBShah:
Cc @MikeInnes

downloads/plotting.md
```diff
@@ -60,6 +60,4 @@ draw(SVG("output.svg", 6inch, 3inch), plot([sin, cos], 0, 25))
 
 Gadfly's interface will be familiar to users of R's
 [ggplot2](http://ggplot2.org) package. See
-[examples](https://github.com/dcjones/Gadfly.jl/tree/master/doc) and
-documentation on the [Gadfly](https://github.com/dcjones/Gadfly.jl)
-homepage.
+examples documentation on the [Gadfly](http://gadflyjl.org) homepage.
```
tkelman: examples and

ChrisRackauckas:
>The post install thing should be a separate PR anyways. Any takers?

I may be interesting in doing it, but this is definitely something that many would like to bikeshed first. Any "recommendations" on the website is going to be highly controversial because it would have to choose some criteria for

- What domains "matter" (why is plotting so important?)
- What packages/orgs should be mentioned (is PyPlot still "widely accepted" enough?)

I have some proposals on this topic for making the guidelines, so I'll open an issue to discuss this before coming up with a concrete PR. But I think this is very important for Julia since people do tend to expect it to "operate out of the box like MATLAB/SciPy/R", and so new users should be somehow pointed (in the least controversial way possible) to the proper resources (which is much a very difficult issue because our package system for even this core functionality is distributed and not centralized to one repo).

ViralBShah:
For history - the plotting thing was a major stumbling block in the early days, and merited a mention. It is still really important - as in people in all domains need to plot, but it almost just works now.

ararslan:
I'll go ahead and merge this and the rearrangement can take place elsewhere as mentioned.

PR #521, opened 2017-03-03
Remove LearnBase from SoC

Evizero:
I am sorry to say, but I just today discovered I am listed as a SoC Mentor which I new nothing about. While I am always happy to help out at Projects I am involved in, I do not think I can promise the time to take on the responsibilities of being an official Mentor. I don't think a promise like this should be made so lightly.

There must have been an internal miscommunication along the line, because both "JuliaML" project ideas listed here are news to me. I am unsure what a SoC-length "LearnBase" project could even entail at this point, so maybe it is better to just remove the idea. I think to remember @tbreloff writing some text like this last year, so maybe it is some leftover text from the past?

I am sorry for this mixup. I am sure the intentions were good. I just don't want to make promises I probably can't keep.

### DIFF
--- a/soc/projects/general.md
+++ b/soc/projects/general.md
@@ -7,20 +7,6 @@ title:  General Projects – Summer of Code
 
 {% include toc.html %}
 
-## Expanding and improving LearnBase/JuliaML
-
-[LearnBase](https://github.com/Evizero/LearnBase.jl) is an attempt to create common abstractions and tools for machine learning frameworks. Common functionality like data loading, loss functions or training processes can be defined generically, and then easily combined with a specific approach (e.g. an SVM).
-
-Possible projects are:
-  * Contributing to the generic functionality in JuliaML (e.g. creating or improving a package for loading data sets, or implementing training process, etc)
-  * Pure-Julia implementations of specific model types (e.g. decision trees or SVMs) which integrate with the JuliaML ecosystem.
-
-**Required Skills**: Knowledge of the machine learning technique(s) to be implemented.
-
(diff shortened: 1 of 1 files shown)


tbreloff:
I've made it very clear that I'm not able to be a mentor, though I do think it can be useful (when properly qualified) to list some potential projects with the clear statement that mentors would need to be recruited.  I think the LearnBase text was from last year.

I'm happy to delete it.

ararslan:
I was in a similar situation having been listed as a mentor. It seems the organizers were pressed for time, since GSoC required potential mentor names to be submitted by February 23rd or so, though they were allowed to change mentors after initial submission. I think in the interest of time it was an "ask forgiveness rather than permission" sort of situation.

ChrisRackauckas:
Related, https://github.com/JuliaLang/julialang.github.com/commit/581d6d441570b3935f7634518c4d3e7678bd9942 .

I think each mentor needs to be checked in with before adding them to the list. There was a deadline where Google wanted some changes done in a day, so I think some of it got rushed. But we should double check that everyone has actually confirmed. I've been doing this by asking each in Gitter or by linking them to the threads, but I am not sure if this has happened for everyone. 

@shashi and @MikeInnes, can we compile a list of who may still need to be notified that they were added, and double check that they will be available?

MikeInnes:
Sorry for the mix-up here. We had a set of people we thought had
volunteered recently, but obviously this got confused in some way, and
we'll remove any names that we need to.

Of course, at no point did we expect to commit people to things they hadn't
agreed to, and the mentors list is meant to be a set of good contacts for
opening issues etc more than anything. In future we'll make sure that these
things are as clear as possible.

PR #540, opened 2017-03-27
Added JuliaDelhi

rahulkp220:
Added juliadelhi's meetup information on line 245.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -242,6 +242,7 @@ Asia
 
 * [Bangalore Julia User Group](http://www.meetup.com/Bangalore-Julia-User-Group/)
 * [Beijing Julia User Group](https://www.meetup.com/juliacn/)
+* [JuliaDelhi](https://www.meetup.com/Juliadelhi/)
 
 Australia
 


rahulkp220:
@ararslan Can we add our Github, Facebook and twitter links as well? Let me know if adding all or any one of these is an option.

ararslan:
I'm not sure what the policy on such links is, if any. @ViralBShah?

ChrisRackauckas:
Those links are at the top of the Meetup page, so is it really necessary to copy them all over?

rahulkp220:
@ChrisRackauckas yeah you are right, but I saw dedicated sections for Github and Twitter, which made me think. However, I am okay if those are reserved for official JuliaLang's repos and links.

ChrisRackauckas:
>However, I am okay if those are reserved for official JuliaLang's repos and links.

Those aren't official. But most of those are only Github orgs, or only Twitter (channels?). I don't think anything else doubles up into multiple categories.

rahulkp220:
I saw juliaLangEs (Mexico) and Julia Tokyo following similar route.

PR #542, opened 2017-03-31
Self-host juliacon 2017 logo

staticfloat:
This should get us the green lock on browsers, without this browsers complain that we're mixing HTTP and HTTPS content, and we can't request over HTTPS from juliacon.org because we don't have that setup yet.  In any case, with HTTP/2 it's faster to have everything on one server anyway, so let's just self-host it.


staticfloat:
screen shot 2017-04-02 at 12 02 27 am

PR #543, opened 2017-04-04
Update index.md

scls19fr:
Add another upcoming Julia book


ViralBShah:
I think it is best to list books once they are already available.

PR #545, opened 2017-04-09
change arm and power links back to 0.5.0

tkelman:
@viralbshah why did you need to wait for me to change 4 characters?


ViralBShah:
Because you do the releases.

tkelman:
This isn't making a release, it's changing a download link.

ViralBShah:
Is something preventing this merge (I only ask because you are writing comments here but not merging, so perhaps there is a reason)? 

Can we please end this extremely silly thread and move on to better things?

PR #561, opened 2017-05-02
fixed build with POMDPs.jl publication

zsunberg:
The POMDPs.jl publication has been in publications/julia.bib, however it does not show up on the site. I think this was because index.md was not rebuilt and there was perhaps an error in the bib formatting? I think this PR should fix it. cc: @etotheipluspi


simonbyrne:
I've updated how publications are stored (see #584). I think I've fixed this but if it is still not displaying correctly let me know.

PR #562, opened 2017-05-04
Update minimum requirement info for macOS to 10.8 in download pages

carlobaldassi:
The macOS binaries don't work correctly on 10.7, see https://github.com/JuliaLang/julia/issues/20651 (which doesn't seem to be about to be fixed any time soon an most likely won't be fixed at all).

Although the binary would still have 10.7+ in the name, at least with this change the information on the website would be correct (I had a puzzled coworker wondering why `Pkg.add` was failing).

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -27,7 +27,7 @@ JuliaBox, all of these plotting packages are pre-installed.
 </tr>
 <tr>
     <th> macOS Package (.dmg) </th>
-    <td colspan="6"> <a href="https://s3.amazonaws.com/julialang/bin/osx/x64/0.5/julia-0.5.1-osx10.7+.dmg">10.7+ 64-bit</a> </td>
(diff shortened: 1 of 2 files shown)


ararslan:
Do they work on Mountain Lion (10.8)?

carlobaldassi:
> Do they work on Mountain Lion (10.8)?

I have no way to test that, however from what I understand the binaries _seemed_ to be working on 10.7 except for the issues with libgit2, so possibly yes?

ararslan:
Okay, good to know. It'd probably be worth having a larger discussion on the minimum macOS version we want to claim to support (cc @staticfloat and @tkelman) but in the meantime this LGTM.

staticfloat:
I would support moving the minimum supported mac version up.  While 10.8 may be the technically lowest we can run modern versions of Julia on, we may want to bump the actually "supported" version up a little higher to 10.9, as that's what our buildbots and tests actually run on.

ViralBShah:
I agree that we may want to go to 10.9, but let's at least make it 10.8. Perhaps we should also adjust that in the build scripts.

PR #568, opened 2017-05-13
Thank fastly

staticfloat:
@ViralBShah They also said they'd like a shoutout over Twitter.


ViralBShah:
Good enough if I give a shout on my twitter account? Not sure who is operating the julialanguage account.

staticfloat:
Yeah, I think that should be fine.  Did we finally get the......ahh....... questionable content off of that account?  If not, we should probably shut it down so it doesn't happen again in the future.

PR #570, opened 2017-05-18
add ucla course

Hua-Zhou:
Add Statistical Computing course at UCLA (biostat m280/biomath m280/stat m280, spring 2017)


ararslan:
Awesome that you're using Julia for those courses!

PR #574, opened 2017-05-26
Add generic linux binary install instructions

### LINKED ISSUE #525: We need to tell people how to add a downloaded generic linux binary to the path
I've just walked two people in a row through doing this on the julialang gitter.
I don't mind, but in the interests of avoiding losing the people who become frustrated and give up without finding IRC or gitter etc.
Maybe we should explain a bit more.

[Transcripts here, if anyone thinks that the questions could be enlightening into the mind of users](https://gitter.im/JuliaLang/julia?at=58bfc1f421d548df2ca83343)

It appears that even having a solid idea of what the Path *is* is not standard knowledge.
So it needs to be explained about symlinking into `/usr/bin/local`.
Like is done in the README.md for building from source.

And that should be done somewhere near the download line.
And/Or in a readme file included in the tarball.

rdeits:
Fixes #525 

Also see https://discourse.julialang.org/t/getting-julia-running-on-atom/3932/5

I hope this is sufficiently clear, and that I've chosen the correct recommended installation steps. I'm happy to revise as needed. 

In addition, I've added direct "help" links from the downloads page for each of the major platform releases and release candidates.

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -25,7 +25,21 @@ Uninstall Julia by deleting Julia.app and the packages directory in ~/.julia. Mu
 
 ## Linux
 
-It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. The following distribution-specific packages are community contributed. They may not use the right versions of Julia dependencies or include important patches that the official binaries ship with. In general, bug reports will only be accepted if they are reproducible on the generic linux binaries on the downloads page.
+It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. 
+
+### Generic Linux Binaries
+
+The generic linux binaries do not require any special installation steps, but you will need to ensure that your system can find the `julia` executable. To do that, you will need to extract the `.tar.gz` file downloaded from the [downloads page](index.html) to a folder on your computer, then either create a symbolic link to `julia` from a folder which is on your system `PATH` or add the `/bin` folder inside that archive to your system's `PATH` environment variable.
+
+As an example, to create a symbolic link to `julia` inside the `/usr/local/bin` folder, you can do the following:
+
+	sudo ln -s <where you extracted the julia archive>/bin/julia /usr/local/bin/julia
+
+On some linux distributions, you may need to use a folder other than `/usr/local/bin`. To check which folders are on your `PATH`, you can run `echo $PATH`. 
+
+### Linux Distribution-Specific Packages
+
+The following distribution-specific packages are community contributed. They may not use the right versions of Julia dependencies or include important patches that the official binaries ship with. In general, bug reports will only be accepted if they are reproducible on the generic linux binaries on the downloads page.
 
 Instructions will be added here as more linux distributions start including julia. If your Linux distribution is not listed here, you should still be able to run julia by building from source. See the [Julia README](https://github.com/JuliaLang/julia/blob/master/README.md) for more detailed information.
 
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -33,7 +33,7 @@ The generic Linux binaries do not require any special installation steps, but yo
 
 * Create a symbolic link to `julia` inside a folder which is on your system `PATH`
 * Add Julia's `bin` folder to your system `PATH` environment variable
-* Invoke the `julia` executable by using it's full path, as in `<where you extracted Julia>/bin/julia`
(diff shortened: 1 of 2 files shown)


downloads/platform.md
```diff
@@ -25,7 +25,21 @@ Uninstall Julia by deleting Julia.app and the packages directory in ~/.julia. Mu
 
 ## Linux
 
-It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. The following distribution-specific packages are community contributed. They may not use the right versions of Julia dependencies or include important patches that the official binaries ship with. In general, bug reports will only be accepted if they are reproducible on the generic linux binaries on the downloads page.
+It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. 
+
+### Generic Linux Binaries
+
+The generic linux binaries do not require any special installation steps, but you will need to ensure that your system can find the `julia` executable. To do that, you will need to extract the `.tar.gz` file downloaded from the [downloads page](index.html) to a folder on your computer, then either create a symbolic link to `julia` from a folder which is on your system `PATH` or add the `/bin` folder inside that archive to your system's `PATH` environment variable.
```
tkelman: bit strongly worded, having julia on the path isn't actually necessary to run it

ViralBShah: Yes, create symlink, or add PATH, or directly invoke the Julia program by the full path (which is what I often do). With this should be good to merge.

rdeits: Good points! I've updated the wording to address this. Could you please take a look?

downloads/platform.md
```diff
@@ -25,7 +25,25 @@ Uninstall Julia by deleting Julia.app and the packages directory in ~/.julia. Mu
...
-It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. The following distribution-specific packages are community contributed. They may not use the right versions of Julia dependencies or include important patches that the official binaries ship with. In general, bug reports will only be accepted if they are reproducible on the generic linux binaries on the downloads page.
+It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. 
+
+### Generic Linux Binaries
+
+The generic Linux binaries do not require any special installation steps, but you will need to ensure that your system can find the `julia` executable. First, extract the `.tar.gz` file downloaded from the [downloads page](index.html) to a folder on your computer. To run Julia, you can do any of the following:
+
+* Create a symbolic link to `julia` inside a folder which is on your system `PATH`
+* Add Julia's `bin` folder to your system `PATH` environment variable
+* Invoke the `julia` executable by using it's full path, as in `<where you extracted Julia>/bin/julia`
```
tkelman: its

also this needs a rebase

downloads/platform.md
```diff
@@ -25,7 +25,25 @@ Uninstall Julia by deleting Julia.app and the packages directory in ~/.julia. Mu
...
+
+For example, to create a symbolic link to `julia` inside the `/usr/local/bin` folder, you can do the following:
+
+	sudo ln -s <where you extracted the julia archive>/bin/julia /usr/local/bin/julia
```
tkelman: should code highlight this

downloads/platform.md
rdeits: Arggggggh. Thank you. Rebase is also incoming.

downloads/platform.md
rdeits: Do you mean bash highlighting specifically? The tab at the start of the line already causes it to be rendered as code on the site, but I don't know how to add shell highlighting inside jekyll.

rdeits: Ah, found it on the homepage source. I'll see if I can bash-highlight all the other snippets on this page.

tkelman: I don't see a tab? I think if 4-space indented at the beginning of a line then markdown would format it as code

rdeits: The highlighter does absolutely nothing for this line in bash-mode, so I'll just leave it indented for generic code formatting

tkelman: it isn't indented though...

tkelman: Hm nevermind, it's visible when i click through to looking at the raw source, but github isn't showing it in the source diff. Maybe 4 spaces instead of a tab would show up properly (tabs are generally annoying and shouldn't be used anyway)

rdeits: Weird. I see it on my end (and in my email notifications). But spaces are better anyway, so I've changed it.

PR #575, opened 2017-05-30
Update index.md

EricForgy:


ViralBShah:
Thanks for pointing out. I am actually going to put the talk schedule up there instead.

PR #578, opened 2017-06-02
install julia with brew on MacOS

osule:
# What does this PR do?
- updates instruction for installing Julia on MacOS

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -23,6 +23,11 @@ On Mac, a `Julia-<version>.dmg` file is provided, which contains Julia.app. Inst
 
 Uninstall Julia by deleting Julia.app and the packages directory in ~/.julia. Multiple Julia.app binaries can co-exist without interfering with each other. If you would also like to remove your preferences files, remove `~/.juliarc.jl`.
 
+An alternative to installing Julia from the downloaded .dmg is by using the [Homebrew package manager](https://brew.sh/). 
+In your terminal, run
+
+    brew cask install julia
+
 ## Linux
 
 It is strongly recommended that the official generic linux binaries from the downloads page be used to install Julia. The following distribution-specific packages are community contributed. They may not use the right versions of Julia dependencies or include important patches that the official binaries ship with. In general, bug reports will only be accepted if they are reproducible on the generic linux binaries on the downloads page.


staticfloat:
@tkelman do we want to officially promote the `cask` version of Julia?  It just downloads and unpacks the official `.dmg`.

tkelman:
Seems a little more predictable than your tap in terms of dependencies? Your tap doesn't have our own copy of libgit2 right, so Pkg breaks any time upstream brew updates libgit2?

staticfloat:
> Your tap doesn't have our own copy of libgit2 right, so Pkg breaks any time upstream brew updates libgit2?

My tap has its problems, but I don't remember this being one of them.  ;)

Mostly I think this should be more robust because it's just the same binary distribution and so if something like the GCC upgrade happens we don't start getting warnings about `libgfortran` missing, etc....

tkelman:
> I don't remember this being one of them. ;)

I do. When libgit2 0.25 came out it was definitely broken.

simonbyrne:
I think we should have this, but:

1. It should state that this is an alternative to downloading the .dmg manually.
2. It also requires the cask tap (i.e. `brew tap caskroom/cask` before).
3. We can also mention the @staticfloat's tap, but emphasise that it involves more compiling and is still slightly experimental (at least until https://github.com/staticfloat/homebrew-julia/issues/231 can be resolved more generally).

tkelman:
cask is enabled by default nowadays IIUC

ViralBShah:
yes. cask is default now.

downloads/platform.md
```diff
@@ -23,6 +23,13 @@ On Mac, a `Julia-<version>.dmg` file is provided, which contains Julia.app. Inst
 
 Uninstall Julia by deleting Julia.app and the packages directory in ~/.julia. Multiple Julia.app binaries can co-exist without interfering with each other. If you would also like to remove your preferences files, remove `~/.juliarc.jl`.
 
+An alternative to installing Julia from the downloaded .dmg is by using the [Homebrew package manager](https://brew.sh/). 
+In your terminal, run
+
+    brew cask install julia
+
+You can also install from the [`staticfloat/julia` tap](https://github.com/staticfloat/homebrew-julia). Note that this tap is still slightly experimental and involves more compiling.
```
ararslan: It shouldn't actually require any compiling though, right? I thought @staticfloat built binaries for his Julia formula regularly so that users wouldn't have to compile anything.

tkelman: not regularly any more

staticfloat:
That tap is essentially unmaintained.  Please don't point people toward it.

downloads/platform.md
ararslan: Ah thanks, didn't know.

osule:
@staticfloat I'm going to remove the sentence pointing to the link. I saw that the tap was unmaintained but wasn't sure if it was going to continue that way. If this however changes in the near future, I think we can always update the documentation.

simonbyrne:
Thanks!

PR #579, opened 2017-06-07
Add the EcoJulia organization

mkborregaard:
We've founded "EcoJulia", an organisation for Ecology in Julia. We're currently two members, Tim Poisot and myself, both university professors in Ecology who use Julia professionally, but we aim to expand. Our first Julia package (SpatialEcology.jl) has just been submitted to METADATA.

https://github.com/EcoJulia

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -86,6 +86,7 @@ Various Julia projects are hosted under the following umbrella organizations on 
 ## Scientific Domains
 
 * [BioJulia](https://github.com/BioJulia) – Biology
+* [EcoJulia](https://github.com/EcoJulia) – Ecology
 * [JuliaAstro](https://github.com/JuliaAstro) – Astronomy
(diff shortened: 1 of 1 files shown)


ViralBShah:
Does it make sense to make it JuliaEco to make it consistent with other org names? Not that it is necessary in any way, and it would entail a bunch of unnecessary work - but just asking once. Feel free to say no.

mkborregaard:
Thanks for asking, I did think about that. I saw somewhere on discourse Stefan Karpinski stating that he saw no reason why all the orgs should start with Julia, but of course that's just a personal opinion.

I am perfectly happy to change it if you prefer, but here are the arguments in favour of keeping this:

1. It is consistent with BioJulia, which it resembles in goal and topic (and hopefully will in scope, eventually).
2. It is easier to say than JuliaEco
3. It leaves the JuliaEcon name open for an economy org - EconJulia is a real tongue-twister.

ViralBShah:
Let's do it!

PR #581, opened 2017-06-12
Include link to aarch64 nightly binary

yuyichao:
I'm not sure what is the expected format so I use the base ISA name so that it's not simply repeating the same information in the parenthesis....

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -121,14 +121,18 @@ Most users are advised to use the latest official release version of Julia, abov
     <td colspan="3"> <a href="https://status.julialang.org/download/osx10.7+">10.8+ 64-bit</a> </td>
 </tr>
 <tr>
-    <th> Generic Linux binaries </th>
-    <td> <a href="https://status.julialang.org/download/linux-i686">32-bit (X86)</a> </td>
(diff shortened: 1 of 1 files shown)


yuyichao:
Comments?

ararslan:
Perhaps a table like this would look more organized?:

| System | Architecture |
|:---:|:---:|
| Windows (.exe) | 32-bit |
|  | 64-bit |
| macOS (.dmg) | 64-bit |
| Linux | 32-bit |
|  | 64-bit |
|  | ARM v7 (32-bit) |
|  | ARM v8 (64-bit) |
|  | PowerPC (64-bit, little endian) |
| RPM Package (.rpm) | 32/64-bit |
| Source | GitHub |

(The cells can be merged vertically as appropriate since that's in HTML, whereas this example is in Markdown.) Anyway, just an idea. What you have here looks fine to me at any rate.

yuyichao:
I think I'll merge this as is for now. I do like the new proposed layout better (though the hardfloat-ness for the armv7 version should be mentioned) but that should probably be done by a HTML expert and also changes the other tables to have a consistent format.

PR #584, opened 2017-06-22
move publication list to jekyll collection

simonbyrne:
This gets rid of the cumbersome bibtex process for adding publications. Now you just add a file to the _publications directory and the result should be formatted nicely.

I don't think we should display the abstracts by default, but I've kept the existing abstract data around in case we want to have some way to display it in future (this is beyond my HTML skills). We could also create a template to generate bibtex as well if we wanted.


simonbyrne:
Also, this gets rid of the "</body> </html>" at the bottom of the current publications page.

ChrisRackauckas:
This looks great!

ViralBShah:
@simonbyrne Merge when ready! We can tweak as we go along.

simonbyrne:
Now live: https://julialang.org/publications/

PR #585, opened 2017-06-24
Add reference for ImageQuilting.jl

juliohm:
@simonbyrne I created the YAML file as you suggested, could you please review and merge?

Thank you.


simonbyrne:
That worked well: https://julialang.org/publications/#HSBC17

If you think it needs more tweaking, let me know.

PR #602, opened 2017-08-17
Hygienic -> unhygienic

mlhetland:
Escaping the code prevents the hygiene mechanism from introducing doing its work, thereby letting you capture variables from the call site – no?

### DIFF
--- a/blog/_posts/2017-08-09-dsl.md
+++ b/blog/_posts/2017-08-09-dsl.md
@@ -314,7 +314,7 @@ function make_model(ex::Expr)
 end
 
 macro model(ex)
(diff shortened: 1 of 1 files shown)


ViralBShah:
@JeffBezanson ?

vtjnash:
Yes, "unhygienic" seems like a better choice of words here. But the example code is also wrong and puts the `esc` around the wrong object. Here's the corrected version:

```
return make_model(make_function(esc(ex)))
```

ViralBShah:
Bump @mlhetland to perhaps update this and get it merged?

mlhetland:
Fixed the use of `esc`, now.

PR #605, opened 2017-08-24
Remove link to PPC nightly binaries

ararslan:
We can add it back once the buildbot is back online, but as of now it's been offline for several months, so there's nothing to download for PPC at the moment.


ViralBShah:
Best not to reinstate when things are better on the ppc front.

PR #607, opened 2017-08-25
Updated DiffEq GSoC Projects

ChrisRackauckas:
Getting ready for the next year, I am sectioning off some projects which would do well as GSoC projects. Some projects are deleted since they aren't as relevant anymore, and projects which are covered by this years GSoC project are also removed. New projects were added.

This of course doesn't need to be merged anytime soon and may change before the next GSoC cycle, but I wanted to get these ideas jotted down.

### DIFF
--- a/soc/projects/diffeq.md
+++ b/soc/projects/diffeq.md
@@ -7,42 +7,32 @@ title:  DiffEq Projects – Summer of Code
 
 {% include toc.html %}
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Happy to merge if you would like to.

ChrisRackauckas:
Sure thing.

PR #610, opened 2017-08-27
Remote outdated references

sathomas:
Neither of the referenced books is expected to be published.


ViralBShah:
Can add them in case they do get published.

PR #618, opened 2017-09-02
FEniCS post rename

ChrisRackauckas:
The FEniCS post wasn't getting added to the blog and so I checked for irregularities and noticed that it doesn't follow the naming structure of the other blog posts, so this PR renames the post in a manner which matches the format.


ChrisRackauckas:
@ysimillides

PR #638, opened 2017-09-27
Clean up the community page

ararslan:
The Git diff is really confusing; I recommend just looking at the rendered page.

This cleans up the community page by making our official channels more explicit and by removing a lot of package-specific information.

If you're wondering why I chose `##` and `###` markdown header levels, clone and build this branch then look at the rendered page; IMO it looks much, much nicer this way, otherwise the headers are enormous.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -7,268 +7,65 @@ All participants in the Julia community are requested to read the
 [Julia Community Standards](standards/), and abide by them in all Julia related
 forums.
 
-# Mailing Lists
+## Official Channels
 
-## Discourse
+### GitHub
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
why not put the Gitter with the IRC? They are at about the same level with Gitter being more trafficked.

ViralBShah:
For meetups, I think just having julia.meetup.com would be sufficient, and the specific ones can be removed. This is far too dynamic and changes a fair bit.

ararslan:
Done.

ViralBShah:
I am merging this and we can improve as we go along.

PR #642, opened 2017-10-09
Update PyPlot example

bcavileer:
sin() and cos() changed to sin.() and cos.()
linestyle is now a symbol

### DIFF
--- a/downloads/plotting.md
+++ b/downloads/plotting.md
@@ -41,8 +41,8 @@ PyPlot.jl and example usage are as follows:
 {% highlight julia %}
 Pkg.add("PyPlot")
 using PyPlot
-x = linspace(0,2*pi,1000); y = sin(3*x + 4*cos(2*x))
-plot(x, y, color="red", linewidth=2.0, linestyle="--")
+x = linspace(0,2*pi,1000); y = sin.(3*x + 4*cos.(2*x))
+plot(x, y, color="red", linewidth=2.0, linestyle=:dash)
 {% endhighlight %}
 
 ## Gadfly


ViralBShah:
The PyPlot.jl README itself still does not use the symbol for linestyle. Is this in a released version of PyPlot.jl?

johnfgibson:
PyPlot 2.3.2 in julia-0.6.0 accepts any of ``"--"``, ``"dashed"``, ``:dashed`` as a value for ``linestyle`` with no deprecation warnings on any of them. I just checked.

Agreed that sin() and cos() should change to sin.() and cos.(); the former do give deprecation warnings.

ChrisRackauckas:
Yes, dots should stay. But I would respect the way the author of PyPlot writes the tutorial, or at least put a comment after that the other is allowed.

ViralBShah:
@stevengj Do you have a preference on the way we do `linestyle` here?

stevengj:
This depends on Matplotlib, not PyPlot: PyPlot just passes this keyword through to Matplotlib without looking at it it.

The [matplotlib documentation](https://matplotlib.org/gallery/lines_bars_and_markers/line_styles_reference.html?highlight=line%20style%20reference) seems to prefer `"--"`, but [`"dashed"` also works](https://matplotlib.org/api/_as_gen/matplotlib.lines.Line2D.html#matplotlib.lines.Line2D).

PyCall converts Julia Symbols into Python strings, so `:dashed` or `Symbol("--")` should also work.  But since the Matplotlib documentation uses strings, I would suggest using a string as well.

ViralBShah:
Ok - clearly makes sense to leave it as a string. Can merge once fixed.

PR #647, opened 2017-10-17
Benchmarks: added Go, abs2 in mandelbrot, moved plot to front page

johnfgibson:
This PR

  * Adds benchmark results from Go 
  * Switches to a uniform user-defined function `abs2` within the mandelbrot benchmark and renames that `userfunc_mandelbrot`. 
  * Folds in revised results from a small `range(t)` fix in Python benchmarks [JuliaLang/julia#24089](https://github.com/JuliaLang/julia/pull/24089)
  * Swaps the placement of the benchmark plot (now on front page) and table (now behind a link)
  * Revises front-page text describing the aim of the benchmark, trying to highlight the focus on identical algorithms and code patterns instead of fastest method of computing the result.


_includes/benchmarks.html
```diff
@@ -1,30 +1,31 @@
 <table class="benchmarks">
 <colgroup>
-<col class="name">
-<col class="relative" span="11">
+<col class="name"></col>
+<col class="relative" span="12"></col>
```
StefanKarpinski: Can we get these without the `</col>` – browsers render them in the output? I made that change to Julia master already, so if you rebase your changes, the Perl script that generates this should not include these.

ViralBShah:
Are all the other benchmarks using `abs2` uniformly. I only see the julia change in here, and am assuming that we may have needed it in other cases too.

KristofferC:
The actual benchmark code is in the Julia repo though, right?

johnfgibson:
@ViralBShah: I'm working on a PR to JuliaLang/julia with the change to `abs2` and name `userfunc_mandelbrot` in all the benchmark codes, probably will file tomorrow.

ViralBShah:
Thank you. Forgot for a moment that these are separate repos.

ChrisRackauckas:
Anything blocking this?

StefanKarpinski:
I wasn't sure if we should merge this before a PR exists for the corresponding code changes, but sure, no harm I guess.

PR #650, opened 2017-10-22
Prune GSoC

ChrisRackauckas:
Get rid of completed GSoC projects to clear up space and make it easier to find what's available. 

@timholy do you agree with taking image segmentation off?
@MikeInnes are there any of the Flux.jl projects to remove?


MikeInnes:
You can get rid of the first two ML projects (torch and optimisers). I'll add some more in when I can.

timholy:
Yes, image segmentation can be taken off.

PR #654, opened 2017-10-26
Update 0.6 download links to 0.6.1

ararslan:
I bless the rains down in Africa

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -3,7 +3,7 @@ layout: default
 title:  Julia Downloads
 ---
 
-# Current Release (v0.6.0)
+# Current Release (v0.6.1)
 
 We provide several ways for you to run Julia:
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Does 0.6.0 go into older releases as part of this?

ararslan:
I'm not sure. I thought we only did that for minor versions rather than patch versions but I don't actually know. I can put 0.6.0 there if you'd like.

ViralBShah:
To answer my own question, it doesn't seem so. Unrelated, We should probably mark 0.5 as unmaintained though.

ararslan:
Don't we typically (at least claim to) support bugfixes for the previous minor version? So when 0.7 is released, 0.5 would be unmaintained and 0.6 would be bugfix-only?

PR #655, opened 2017-10-27
Move all macOS download URLs to new location

ararslan:
As part of the process of normalizing the structure of the URLs on AWS, we're renaming `osx` to `mac` everywhere, and replacing `osx10.x+` in the file names to `mac64`. **All links to the old URLs will still work; the files in the `mac` tree are copies of those in the `osx` tree, which still exists.**

Note that with this new naming scheme, OS X 10.6 and 10.7+ are no longer distinguishable in a consistent manner. Because of that, and because both OS X 10.6 and Julia 0.2 are _way_ past EOL, the link for Julia 0.2 on OS X 10.6 has been removed. The new link points to the download for 10.7+.

This PR also moves the downloads for Julia 0.1 from Google Drive to AWS.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -27,7 +27,7 @@ JuliaBox, all of these plotting packages are pre-installed.
 </tr>
 <tr>
     <th> macOS Package (.dmg) <a href="platform.html#macos">[help]</a></th>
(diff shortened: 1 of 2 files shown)


ararslan:
Thanks for catching that @staticfloat!

PR #657, opened 2017-10-28
Move all the benchmarks stuff into the benchmarks/ directory.

ViralBShah:
Move some of the benchmarks details also into the benchmarks page.


ViralBShah:
cc @johnfgibson

PR #662, opened 2017-11-07
Do not MERGE: 3.4.6 vs. "Anaconda Python v3.5.4"

PallHaraldsson:
The table seems inconsistent saying "3.4.6"; in case it is wrong (probably, I just didn't clearly see how to change the table), it should be synchronized.

Sync either way that is correct, or even better there's Python 3.6.x out, currently newest, and use it and matching Anaconda.


johnfgibson:
Good catch, thanks. The error is in the table, which was produced with a different python3 in my PATH than anaconda python. I'll just have to make sure to run the table in the same environment as the benchmarks runs. 

The current Anaconda version is listed as 3.6 on https://www.anaconda.com/download/#linux, but when you download the binary for Linux, it's actually 3.5.

johnfgibson:
This PR should be closed. The issue was fixed in [#655](https://github.com/JuliaLang/julialang.github.com/pull/665)

PR #678, opened 2017-11-20
Rename JuliaStats URL

Ismael-VC:
Rference:

* https://discourse.julialang.org/t/http-juliastats-org-domain-seems-to-have-expired/7192


ararslan:
I forgot we even had a site like that for JuliaStats.

PR #682, opened 2017-12-04
Blog Formatting

MikeInnes:
Not a drastic change, but makes various small things more readable. The previous wide column format really felt like a huge wall of text. Also adds quote formatting, inline code formatting, and a `desc` class for inline image captions.

screenshot 2017-12-04 16 56 17
screenshot 2017-12-04 16 56 31
screenshot 2017-12-04 19 12 06

### DIFF
--- a/css/screen.css
+++ b/css/screen.css
@@ -215,11 +215,12 @@ code, table.benchmarks tbody {
 
 pre {
   border: 1px solid #ddd;
-  margin: 1em 0;
-  padding: .75em 1.25em;
-  line-height: 1em;
+  background: rgb(250, 250, 250);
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Not the biggest fan of large margins, but I like everything else.

ViralBShah:
I usually prefer narrow columns of text, but I just resize my browser.  I'd be ok with this.

ararslan:
I think this looks much nicer. Good work, Mike!

ViralBShah:
Let's merge it for now. Can always revert it if we change our minds or don't like it or get too many complaints.

PR #685, opened 2017-12-07
Update benchmarks.html

waldyrious:
This PR syncs the generated html with the source code that produces it.
See https://github.com/JuliaLang/julia/pull/24565 for context.

The changes in this PR include:

- whitespace adjustments, for source readability and consistency
- replacing a hard hyphen in "Mathematica" with a soft hyphen (`&shy;`)
- changing the first cell in the header from a `<td>` to a `<th>`, to match the other ones
- adding a comment at the top of the file to indicate its source.

Of these, only the hyphen modification will produce changes in the rendered html.

However, two other changes are included here:

- The Matlab column was moved 2 places to the right
- The `--` in Javascript's `print_to_file` benchmark was replaced with a `0.00`.

@johnfgibson can you confirm which of the latter changes, if any, are correct, and whether the generating script would need any adjustment in light of this?


ViralBShah:
Bump

waldyrious:
@ViralBShah I'm assuming you meant to ping @johnfgibson, right?

johnfgibson:
Ah, thanks for the ping & bump. Will check it out at work tomorrow.

johnfgibson:
The Matlab column should stay where it was, between JavaScript and Mathematica, as it is in the benchmark plot. The languages are listed in order of geometric mean of the microbenchmarks. Matlab briefly jumped two columns to the right due to a bug (a missing semicolon generated loads of extraneous text processing and poor performance of `userfunc_mandelbrot` https://github.com/JuliaLang/julialang.github.com/pull/656), but when this was fixed it regained its place. 

(The geometric-mean ordering is computed by the `benchmarks.ipynb` notebook, which then produces the `benchmarks.svg` figure with the correct ordering. When changes have occured I've then gone and reordered the languages in `table.pl` to match. Looks like I missed switching Matlab back, or we got out of sync.)

The `-- `in JavaScript's `print_to_file` is meant to indicate a missing value --JavaScript is sandboxed and can't print to a file, as far as I understand. The perl script as is generates 0.0 for this missing datapoint. I've been changing that to `--` manually. IMHO it'd be nice if the table.pl did this automatically, but I don't think it's a big deal either way.

It sure is nice to have the HTML formatted legibly!

waldyrious:
@johnfgibson thanks for the clarifications. I've directly edited this PR to make those changes. The reordering (and replacing `0.00` with `--`) shouldn't be hard to do on `table.pl`, but I'd rather not touch it again after messed up the last time I edited it :sweat_smile:

Issue #549, opened 2017-04-11
State of website translations


ViralBShah:
The language translations dropdown needs some work. I think it should probably be sorted by common languages and level of completeness. Also, the full list no longer shows because of the sheer number of translations.

In many cases, when I click on the translation, I almost always see the first paragraph in English, which probably would make someone think that the translation is incomplete.

Cc @Ismael-VC

waldyrious:
@ViralBShah I started some work related to the dropdown in #551 (as discussed in #428); please review when you can.

ViralBShah:
I wonder how effective the whole translation thing is. There are tons of languages, but on most of them, I just get English. Perhaps we should only show those which have at least 75% complete on the landing page?

Ismael-VC:
@ViralBShah I'm on it!

Ismael-VC:
@ViralBShah done, I'll promote translators and reviewers to complete the home page first.

![image](https://cloud.githubusercontent.com/assets/4594825/26032357/2651e276-3881-11e7-8e31-33c8c3b3c450.png)

Ismael-VC:
These are the languages with at least 75% of translation and reviewing in the home page, many don't show the first paragraph translated, since it has changed and it needs to be reviewed again. I'll make everyone know about this.

ViralBShah:
Ah, I see. 75% seems like a good bar. We can then have a link in that drop down for other languages where more translation is required.

jebej:
Where are these translations? I could contribute.

Ismael-VC:
@jebej the translation project is at:

* https://www.transifex.com/julialang-i18n/julialang-web

Thanks for your support! 👍

ViralBShah:
Looks like we only have English and French now. Perhaps this translation stuff is not quite working out. How about we simply let Google translate take care of it?

waldyrious:
> Perhaps this translation stuff is not quite working out.

@ViralBShah sorry for the long delay -- I just submitted #686. I trust that more people will start contributing once the prompt is presented in a more visible way. Let's give it a few months and see whether it improves the rate of translations.

PR #686, opened 2017-12-07
Add translation prompt to transifex dropdown

waldyrious:
Follow-up of #551; addresses #549. /cc @Ismael-VC and @ViralBShah (sorry for the long delay!)

Note: the bulk of the changes in the diff are cleanup. The main commit for this change is cd13fd5.

### DIFF
--- a/_layouts/common.html
+++ b/_layouts/common.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en-us" lang="en-us">
+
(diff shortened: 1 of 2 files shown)


ViralBShah:
Let's see how this works out.

Ismael-VC:
I'm getting:

```javascript
Uncaught TypeError: Cannot read property 'appendChild' of null
    at (index):1072
```

Ismael-VC:
https://github.com/JuliaLang/julialang.github.com/pull/687 fixes the error above.

PR #687, opened 2017-12-08
Fix wrong id for transifex widget.

Ismael-VC:
Fixes:

```js
Uncaught TypeError: Cannot read property 'appendChild' of null
    at (index):1072
```

### DIFF
--- a/_layouts/common.html
+++ b/_layouts/common.html
@@ -46,7 +46,7 @@
       txPrompt.appendChild( txPromptLink );
 
       // Add the translation prompt list item to the list of languages
-      document.getElementById( "tx-live-lang-picker" ).appendChild( txPrompt );
+      document.getElementById( "tx-live-lang-toggle" ).appendChild( txPrompt );
     </script>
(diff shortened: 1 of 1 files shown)


waldyrious:
Did you successfully test this? If I try it, it adds the link to the menu header, rather than as an entry in the list of languages.

I'm pretty sure the "tx-live-lang-picker" ID is the right one, but the script must wait until it is created dynamically before it runs. I'll see if I can introduce some sort of listener.

waldyrious:
After some time reading about Mutation Observers. More than I would have liked, actually -- I ended up modifying the [MDN example](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#Example_usage) because, like everything else I was finding online, it included unnecessary complexity that made the basic concept harder to grasp.

Anyway, it looks like a possible way to fix this, although quite verbose (it will pretty much double the size of this script). I'll submit a PR with the proposed fix; hopefully it'll work out.

PR #688, opened 2017-12-08
Use MutationObserver API to avoid a null access error

waldyrious:
#686 had a flaw, in that if the script ran before the transifex dropdown had been instantiated, it would attempt to access a null element. This change should fix that.

See #687 for additional context. /cc @Ismael-VC and @ViralBShah.

### DIFF
--- a/_layouts/common.html
+++ b/_layouts/common.html
@@ -45,8 +45,26 @@
       var txPrompt = document.createElement( "li" );
       txPrompt.appendChild( txPromptLink );
 
-      // Add the translation prompt list item to the list of languages
-      document.getElementById( "tx-live-lang-picker" ).appendChild( txPrompt );
+      // Callback function to execute when mutations are observed
+      var callback = function(mutationsList, observer) {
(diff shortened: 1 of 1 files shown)


ViralBShah:
So, which ones to merge?

waldyrious:
Perhaps @Ismael-VC can confirm whether the assertions I made on #687 are right. My guess is that swapping the ID doesn't solve the fundamental issue (and actually adds the link to the wrong place), but I could be missing something.

waldyrious:
@ViralBShah sadly @Ismael-VC has given up on his patch and apparently on the i18n of Julia altogether. I'd suggest to merge this one and see if it works. We can always revert it if it doesn't.

waldyrious:
Argh! It didn't work :( feel free to revert the commit, @ViralBShah, and in the meantime, I'd appreciate if you could ping someone who you believe might know how to correctly fix this. My experience with web development is pretty much limited to small experiments here and there.

PR #701, opened 2018-01-08
Benchmark update for bugfix in python printfd

johnfgibson:
Updated benchmark data & plot for python `printfd` bugfix, https://github.com/JuliaLang/julia/pull/25456.

Several languages see factor 2-ish changes in their `print_to_file` benchmarks due to standardizing on formatting `i, i+1`. Previously some languages formatted `i, i`.


ViralBShah:
Do we need to update the svg on the julia front page as well?

PR #718, opened 2018-01-22
Include information about building system image

AustinPrivett:
I have been using Julia for a couple of months now and did not come across http://www.stochasticlifestyle.com/7-julia-gotchas-handle/ until recently. 

If building the system image still does what is mentioned at that website (i.e., allow all compiler optimizations unlike the pre-built binaries), I think it is helpful to point that out early on to give new users better performance.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -59,6 +59,12 @@ trouble installing Julia.  Checksums for this release are available in both [MD5
 If the provided download files do not work for you, please [file an
 issue in the Julia project](https://github.com/JuliaLang/julia/issues).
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
I feel like this is something that belongs somewhere in the Julia docs - but not directly on the downloads page. I wonder what others think.

ararslan:
I agree. To me this is more "advanced user" territory that 99% of users shouldn't need to care about. @AustinPrivett, would you be willing to submit a PR to Julia to add this information to the documentation?

fredrikekre:
See also https://github.com/JuliaLang/julia/issues/24646

KristofferC:
For Julia 0.7 this is not accurate anymore (pre built binaries are built for multiple targets and jit compiled code is compiled for native target).

AustinPrivett:
In light of the modifications to Julia 0.7 mentioned by KristofferC, I'm closing this and not worrying about adding this information to the documentation.

PR #724, opened 2018-02-02
adding JuliaData as ecosystem

mbesancon:
added the JuliaData org to the list of ecosystems in the Data category


ararslan:
👍 Thanks!

PR #726, opened 2018-02-11
Develop an AI package in Julia based on AIMA

sambitdash:
AI requires well defined definitions to be effectively develop applications. AIMA frameworks can benefit practitioners in designing their problems better. Such a framework is not offered by any other libraries today.

### DIFF
--- a/soc/projects/ml.md
+++ b/soc/projects/ml.md
@@ -144,3 +144,88 @@ and classifying when specific approaches are appropriate will be developed
 simultaneously.
 
 **Mentors**: [Chris Rackauckas](https://github.com/ChrisRackauckas)
+
+## Artificial Intelligence Library Package based on Artificial Intelligence - A Modern Approach (AIMA)
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This is far too detailed. It should be just a short paragraph summary. Also this needs mentors. @sambitdash are you agreeing to mentor?

ChrisRackauckas:
Also, this is written in the first person "an attempt has been made by me". It repeats a lot of things already written into the application materials like how much Julia knowledge is required, and just doesn't match the format of the others. This should get reverted and reworked to be like the other projects.

ViralBShah:
Shorter as discussed and avoiding first person would be nice to have, but I prefer not to revert. I do agree about the mentorship.

@sambitdash Would you be willing to mentor?

sambitdash:
Ok let me fix the write up and the mentor section and resubmit with another PR. I can be the mentor.  But if there is someone more experienced interested happy to delegate as well.

PR #727, opened 2018-02-12
Update ml.md to shorten the AIMA project

sambitdash:
If there is further trimming required will be happy to do so.

Refer to https://github.com/JuliaLang/julialang.github.com/pull/726 for comments. 

@ChrisRackauckas, pls. review the content.

### DIFF
--- a/soc/projects/ml.md
+++ b/soc/projects/ml.md
@@ -147,45 +147,18 @@ simultaneously.
 
 ## Artificial Intelligence Library Package based on Artificial Intelligence - A Modern Approach (AIMA)
 
-AIMA is a seminal text on representation of agents that can aid solve a problem for 
-an environment of varying complexity. Contributed by industry and academicians in 
-general has enriched its content over past two decades and many of the proposed 
-architectures are vetted time and again. The book has updated several times and each 
-revision kind of presents itself as a significant rewrite with all modern content. 
-The book is almost read by every AI engineer as a text for their graduation studies.
-So most people can connect to the content very easily. An AI framework developed 
-around the examples will provide a good foundation to design one's problem that can 
-utilize various non-ML or ML based solvers. 
+AIMA is a seminal text on representation of agents to solve AI problems. Most packages 
+available today as AI libraries tend to focus on ML only and not the architectural
+aspect of AI. The scope of the project is to create a library with a clean API 
+(following AIMA) to easily allow the application of core algorithms to AI problems. 
+The student will implement a package that brings together implementations of algorithms 
+like depth-first search and simulated annealing, both from other Julia packages and 
+from sample code in the AIMA book, and build sample programs to demonstrate AI 
+applications. Starter code can be found at AIMACore along with AIMASamples.
 
-Most packages available today as AI libraries tend to focus on ML only and not the
-architectural aspect of AI. Very few will talk about representational aspects of AI. 
-Julia with abstracted type systems and multiple dispatch interfaces provide an ideal 
-language to represent the problem abstractions best as strategy patterns of software 
-engineering. So you can write the AIMA algorithms very easily as method with no side 
-effects in generic form. Julia provides connectivity to other languages for using their 
-libraries. Thus makes an ideal choice to represent the framework while can use solvers 
-from other languages. While AIMA authors have developed the sample code, their focus 
-is more towards developing sample code for teaching and learning and not professional 
-usage of algorithms.
+**Recommended Skills**: Previous experience with AI or the ability to quickly pick up on 
+the AI algorithms in AIMA
 
-**Deliverables**
-
-The scope of the project is to create a library of core algorithms and sample programs 
-separated out neatly such that core algorithms can be used by practitioners for abstract
-problem representations. 
-
-An attempt has been made: [AIMACore](https://github.com/sambitdash/AIMACore.jl) and 
-[AIMASamples](https://github.com/sambitdash/AIMASamples.jl). `AIMACore` can easily be 
-used by a practitioner in their production projects while `AIMASample` can provide examples
-or testcases for the `AIMACore`. The current code has tried to address all the search 
-problems cases. There is a need to extend it to the rest of AIMA sample code. 
-
-**Skillset**
-
-1. A person should be knowledgable in Julia programming or ready to learn it quickly. 
-2. Should have appreciation for software engineering principles. A framework requires a 
-lot on architectural skills.
-3. Ideally, has either undergone a course in AI or wants to invest time in reading and 
-applying AIMA. 
+**Expected Results**: A well-documented library of functions derived from the AIMA book.
 
 **Mentors** [Sambit Kumar Dash](https://github.com/sambitdash). 
--- a/soc/projects/ml.md
+++ b/soc/projects/ml.md
@@ -147,27 +147,7 @@ simultaneously.
 
 ## Artificial Intelligence Library Package based on Artificial Intelligence - A Modern Approach (AIMA)
 
-Often AI and ML are considered synonymous fields, while in most cases the scope 
-and focus is entirely different. AI is a field of identification of intelligent 
-problems and developing representational frameworks which can aid arriving at a 
-solution better. AI establishes or ensures you set up the problem domain properly.
-Apply disparate solution schemes using existing technologies based on determistic 
-methods or non-deterministic methods to arrive at a solution. The scope of AI is 
-establishing the architecture for solution and ML is to actually solve the problem.
-Not all problems can be solved by all ML techniques known. For someone to be able 
-to try out various schemes to the similar problems may end up rearchitect the 
-problem again for new algorithm. Two examples below can provide concrete 
-visualization to the claims. 
-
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
It really doesn't need to be all that much. Just one maybe two paragraph into. A few lines for expected results, a few lines for recommended skills. The rest of the details should be tailored to the student via discussions with the mentors and the community. Take some of the others as examples, like:

https://github.com/JuliaLang/julialang.github.com/blob/master/soc/projects/compiler.md#thread-safety

ChrisRackauckas:
IMO it seems like there's two concrete projects there: an information abstraction algorithm and algorithms for search problems. A good approach to condense this is to write specifically for those two separately.

1. What is the information abstraction problem? What does it solve? What algorithm is used? How would it be built? What kind of test should it pass at the end of the summer?
2. What is the search problem? What does it solve? What algorithm is used? How would it be built? What kind of test should it pass at the end of the summer?

soc/projects/ml.md
```diff
@@ -147,53 +147,32 @@ simultaneously.
...
-developing sample code for teaching and learning and not professional usage of algorithms.
+effects in generic form. Julia provides connectivity to other languages for using their 
+libraries. Thus makes an ideal choice to represent the framework while can use solvers 
+from other languages. While AIMA authors have developed the sample code, their focus 
+is more towards developing sample code for teaching and learning and not professional 
+usage of algorithms.
 
 **Deliverables**
 
 The scope of the project is to create a library of core algorithms and sample programs
```
ChrisRackauckas: The deliverables usually aren't listed here since they are agreed upon directly between mentor and student given the students background and interests.

sambitdash:
No the intent is to just take all the sample code and make them into a library with proper abstractions.

 The problems proposed were just examples to distinguish AI and ML.

ChrisRackauckas:
Then I have misunderstood what the proposed project is. What algorithms are you specifically asking the student to implement? Note that normally these are undergraduate students (with a few graduate students), so it's good to have a narrow scope and choose a specific thing to do rather than leave it open ended.

sambitdash:
I am going chapter wise, initial chapters have more sample code and later chapters have lesser sample code. So guess chapter 4-12 will be a good target for a month's activity. Is that what the duration of engagement is?
 
The intent is to preserve the sample code as much as practicable to the details to which it is described in the book. No own interpretation is needed for the core algorithm. Make it a method without side effects. 

When you implement the example or sample code you may use data structures or algorithms from Julia's standard libraries or take custom implementations. And pass all such DS as parameters to functions such that the parameters can be changed for another implementation. Example: One may use a queue from standard data structure but the Queue could have been a database query supported by an SQL in the backend. Essentially, code should accept any DS which supports the queue trait. 
 
For example: In `UniformCosSearch` [technique]( https://github.com/sambitdash/AIMACore.jl/blob/56b15f3b1f95b194bb06bbc15ca1887c75a6fa67/src/search.jl#L205),
You can use any Priority Queue, Set or State objects as long as the traits are validated in the [constructor](
https://github.com/sambitdash/AIMACore.jl/blob/56b15f3b1f95b194bb06bbc15ca1887c75a6fa67/src/search.jl#L218)

[`execute`](https://github.com/sambitdash/AIMACore.jl/blob/56b15f3b1f95b194bb06bbc15ca1887c75a6fa67/src/search.jl#L228)  is a set of commands which is almost a verbatim representation of the code in the book.

ChrisRackauckas:
I took your text and cut out the parts which are not related to SoC. Here's what I came up with:

AIMA is a seminal text on representation of agents that (can aid solve a problem for an environment of varying complexity ? *That needs work*). Most packages available today as AI libraries tend to focus on ML only and not the architectural aspect of AI. Very few will talk about representational aspects of AI. The scope of the project is to create a library of core algorithms and sample programs separated out neatly such that core algorithms can be used by practitioners for abstract problem representations. (* What algorithms? What's needs to be done? *)

Recommended Skills: previous experience with AI or the ability to quickly pick up on the AI algorithms in AIMA 

Expected Results: ?


Hopefully that template puts us back on the right track.

sambitdash:
Let me clarify a bit further: 

1. This is a software engineering project not a pure coding project not a research or scientific project. 
2. If you look at the [N-Queens Problem](https://github.com/sambitdash/AIMASamples.jl/blob/master/src/Queens.jl) you can see the same problem definition can use the following search algorithm just changing the `SearchAlgorithms`:
          a. GraphSearchAlgorithms
             1. Depth First
             2. Breadth First
             3. Iterative Depth First
             4. A* algorithms
          b. Genetic Algorithm
          c. Simulated Annealing
3. You can do so very easily as you are following the interface definition of a `Problem`, `SearchAlgorithm` verbatim as described in the book. 
4. These abstractions are very useful for an engineering practitioner. 

The same interfaces can be used to solve real life applications as well if the abstractions are kept pure. 

The intent of the project is to take all such algorithms described in the book as called out sections (if you look at the book those are clearly placed in well outlined sections) and reproduce them in Julia (it's so well written almost all of them can be described in Julia as is) so that some one using those can use the same alogorithms for production applications.

ChrisRackauckas:
Here's suggested text given the statement above. Let's discuss it from here:


AIMA is a seminal text on representation of agents to solve AI problems. Most packages available today as AI libraries tend to focus on ML only and not the architectural aspect of AI. The scope of the project is to create a library of core algorithms and sample programs using the AIMA example code. The student will implement and optimize Julia versions of algorithms like depth-first search and simulated annealing. These implementations will be tested in real-world applications. Starter code can be found at [AIMACore](https://github.com/sambitdash/AIMACore.jl) along with [AIMASamples](https://github.com/sambitdash/AIMASamples.jl).

Recommended Skills: Previous experience with AI or the ability to quickly pick up on the AI algorithms in AIMA

Expected Results: A well-documented library of functions derived from the AIMA book.

ChrisRackauckas:
Separate from the question of the text, I'm curious about the why. Why are these implementations of graph search algorithms better than the LightGraphs versions to justify a second library? Why are these implementations of optimization algorithms better than the Optim and BlackBoxOptim versions to justify another library? I don't really understand with the examples you've given how to convey to the student that they will be adding a unique and useful feature to the Julia ecosystem that can/will gain adoption.

sambitdash:
These are best interfaces ever designed for AI. It's the interfaces that I am interested in. Not the implementations. Change the interfaces of Optim or all the packages to support the interfaces defined in this book. They will be perfect for addressing larger classes of AI problem. Let me say this is the STL or boost of AI. People love those libraries for their usage ease and variety of problems they solve. Not to address one critical application of data structure to their problem set.

sambitdash:
@ChrisRackauckas, find one implementation in the industry or code anywhere just mere changing of the algorithm you can use the same code to use graph search vs. genetic algorithm. That is the power of the interfaces or framework. That makes code easy to use or later to maintain.

ChrisRackauckas:
>These are best interfaces ever designed for AI. It's the interfaces that I am interested in. Not the implementations. Change the interfaces of Optim or all the packages to support the interfaces defined in this book. They will be perfect for addressing larger classes of AI problem.

So it's about the interface and not about the implementations of the algorithms? That wasn't mentioned in the previous texts and would be good to highlight if that's the case.

ChrisRackauckas:
How's this then?

AIMA is a seminal text on representation of agents to solve AI problems. Most packages available today as AI libraries tend to focus on ML only and not the architectural aspect of AI. The scope of the project is to create a library with a clean API (following AIMA) to easily allow the application of core algorithms to AI problems. The student will implement a package that brings together implementations of algorithms like depth-first search and simulated annealing, both from other Julia packages and from sample code in the AIMA book, and build sample programs to demonstrate AI applications. Starter code can be found at AIMACore along with AIMASamples.

Recommended Skills: Previous experience with AI or the ability to quickly pick up on the AI algorithms in AIMA

Expected Results: A well-documented library of functions derived from the AIMA book.

sambitdash:
"both from other Julia packages and from sample code"
 
From other Julia packages, immediately will be a tall order for now. I was trying to do the same for Genetic Algorithm for my problem set and found a lot of effort in doing so. For now it may be just enough to implement the algos verbatim from the book. The next phase will be to impress upon package designers to provide interfaces for this package or implement connectors to other packages from this package if there is adoption. 

The text as you have written definitely conveys the core intent of the project to certain extent. Please, go ahead with your text..

ChrisRackauckas:
Well the authors of many packages, including LightGraphs and Optim, are GSoC mentors. So if it's hard to use them to solve AI problems, then massaging them to an API that makes it easier would be a great GSoC.

ChrisRackauckas:
Let's get someone else to double check.

ChrisRackauckas:
@ViralBShah can we get a second pair of eyes on this. I made some big changes so I want to make sure someone else agrees before merging. Thanks.

ViralBShah:
Thanks. @sambitdash If you want to stick the longer description into a gist and link it here, we can do that too - that way we have the best of both worlds.

sambitdash:
Will do.

PR #734, opened 2018-02-18
Add and update GSOC projects for numerical maths

haampie:
@andreasnoack and @jiahao: maybe you want to mentor these projects as well?


ViralBShah:
Also cc @andreasnoack

Issue #683, opened 2017-12-05
Donate button looks disabled


MikeInnes:
The donate button is greyed out with no hover effect, so it looks like it's not clickable for some reason.

ViralBShah:
I think it is whatever NumFocus provided us. Good to fix it if possible.

ararslan:
It does seem to work, it just looks like it's disabled. We might be able to change the color of it or something.

ihnorton:
https://github.com/JuliaLang/julialang.github.com/blob/master/_layouts/common.html#L180

ViralBShah:
Fixed in the new website.

PR #748, opened 2018-03-02
Make the "Donate Now" button not look disabled

amellnik:
The current color scheme is nice and unobtrusive, but it makes the button look like it's greyed-out and disabled.  This keeps it neutral but adds a color similar to the links on the page to make it clear that it can be clicked on.  Some discussion here: https://github.com/JuliaLang/julialang.github.com/issues/683

image


xorJane:
This looks nice. Thanks Alex!

PR #749, opened 2018-03-04
update list of courses taught at SGH

bkamins:


xorJane:
Thanks for adding those classes!

PR #759, opened 2018-03-13
2018 Pi day post by Cormullion

simonbyrne:
I was a bit tardy, and didn't get around to drumming up more content or writing up a proper post, but fortunately @cormullion has done a beautiful one. 

@cormullion: Is this okay? I realise that this is a bit different than what I proposed, so if you would prefer we didn't put it on the blog, that's fine.


cormullion:
Go for it! 😂😂😂😂

simonbyrne:
Great thanks! I'll wait until we're into 14th March UTC to pull the trigger on it.

PR #760, opened 2018-03-14
Fix P symbol to be phonogram or phonorecord (not published) symbol for…

mbadros:
… clarity when it refers to sound recordings


simonbyrne:
Thanks!

Issue #764, opened 2018-03-16
Structuring of list of classes using Julia for teaching


crstnbr:
The [list of classes using Julia](https://julialang.org/teaching/) has reached quite some length and should probably be structured in some way.

- AFAICT the list has no order. Probably we should just sort it lexicographically.
- It would be cool to have a (interactive) world map visualization of where Julia is being used for teaching.

ViralBShah:
It's going to keep getting longer, so I'd rather just keep it as is for now. But a map seems like a good idea, and you can click to see. Also this list is based on submissions, and is probably quite incomplete. For example, on JuliaBox, we see logins from 600 universities (although not all may necessarily be using Julia for teaching classes).

crstnbr:
I don't see how "it's going to keep getting longer" is a valid point against ordering. One just hast to check that new PR's don't break the order, right?

Regarding a map, I just hacked a quick and dirty version: https://github.com/crstnbr/juliamap

Suggestions and PR's are of course very welcome.

crstnbr:
> Also this list is based on submissions, and is probably quite incomplete.

It's probably important to make it easier for people to add their institution to the list. I'd assume not everyone is familiar with PR's .

ViralBShah:
You can generally just edit the file right here on github. Also, you are right. Let's try some new orderings.

ViralBShah:
The map is pretty awesome. Can we embed it on the teaching/learning page? When you hover on a particular location, could it list the name of the university and the class?

ChrisRackauckas:
Yeah, the map needs some information was taught at the school.

crstnbr:
> You can generally just edit the file right here on github. Also, you are right. Let's try some new orderings.

Right, you can. But it says on the website: _"If you know of other classes using Julia for teaching, please submit a pull request to update this list."_ and it links to the PR page, not the file edit page. Maybe we should change the link.

But why not even add, "or write an email to teaching@julialang.org" (to be replaced by any meaningful email address) or similar to be safe? I'd assume there won't be many mails at a time.

Regarding the ordering, I'll make a PR that sorts the list lexographically soon.

crstnbr:
> The map is pretty awesome. Can we embed it on the teaching/learning page?

Of course. I don't have docker though so I can't really clone the site and make and test the changes.

> When you hover on a particular location, could it list the name of the university and the class?

Makes sense. I'll update it asap.

crstnbr:
[...]

PR #766, opened 2018-03-19
Classes using Julia for teaching: link directly to edit page

crstnbr:
See #764


xorJane:
Thanks Carsten! This is much more clear with your edit.

PR #767, opened 2018-03-19
Add GUI category and QML project

barche:
I also moved the WebIO.jl project to the GUI category, since it doesn't really seem to fit into the IDE category.
@shashi Would you like to co-mentor this, given the link with Observables?


shashi:
I could probably help with the observables part of the work, and only that part :-) co-mentoring would be pretty hard for me. Thanks!

barche:
Yes, that is actually exactly what I meant, thanks :)

PR #768, opened 2018-03-20
Classes using Julia: lexographical ordering and geo information

crstnbr:
See #764. This PR does the following:

* merges multiple separate entries of same institution
* sorts lexographically by institution name
* puts time information (i.e. "Fall 2015") on course rather than institution
* adds (approximate) geo location (latitude and longitude) to institutions in the form of invisible html comments. This is a preparation for integrating the interactive map (see http://github.com/crstnbr/juliamap).

I hope the html comments containing the geo information are really hidden. Github preview says they are but I couldn't check because I don't have Docker available.

### DIFF
--- a/teaching/index.md
+++ b/teaching/index.md
@@ -19,38 +19,56 @@ a
 
 Julia is now being used in several universities and online courses. If you know of other classes using Julia for teaching, please consider [updating this list](https://github.com/JuliaLang/julialang.github.com/edit/master/teaching/index.md).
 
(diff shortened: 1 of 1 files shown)


crstnbr:
Of course, it's also fine to have entries with no geo information. They will end up as `missing` and will be ignored in plotting the map (see https://github.com/crstnbr/juliamap/blob/master/juliamap.jl)

crstnbr:
Feels good to me :)

PR #772, opened 2018-04-05
Add Optim.jl JOSS article

anriseth:
We have submitted a JOSS article on Optim.jl that has just been published, and hope you can add it to the list of publications on julialang.

Should the title have brackets around Julia as in BiBTeX?

Ref @pkofod


ViralBShah:
I don't think it should have the curly braces.

anriseth:
Thanks:)

PR #785, opened 2018-04-25
Update index.md: Think Julia

cormullion:


ViralBShah:
Says `Think Python`

KristofferC:
It says "Julia version of... Think Python" though.

cormullion:
I thought it best to (also) credit the original on which Ben’s work is based...?

PR #788, opened 2018-05-02
Improve consistancy and fix typos

oxinabox:
These typos have been bugging me for while
Also while I was here I made  some of the internationalization resource uses similar descriptions because they have similar content.


ViralBShah:
Do we have an active JuliaML gitter? Should we direct people to use the Julia slack?

oxinabox:
There is a general notice at the top of the page saying that most of these groups have a channel on the slack and a subforum on discourse.
Can we directly link to a slack channel? I think we cannot? I guess we can do the same as for QQ and just give the channel name.

The JuliaML gitter has a few active people. it is much quieter than the slack through, you're correct.

PR #792, opened 2018-05-13
Clarify the historical use of hacks

yurivish:
The current phrasing doesn't make it quite as clear that the hacks used by ML libraries to opt out of fusion are rendered unnecessary by the improved broadcasting machinery.


ViralBShah:
cc @MikeInnes

PR #806, opened 2018-06-25
0.7 alpha -> beta

ararslan:
Everything is uploaded and ready, so this is good to merge.


ViralBShah:
Are the freebsd binaries x86-64 or x86? The binaries have x86_64 in their name, so perhaps they are x86-64 only.

ararslan:
Yes, they are x86-64 only. I just used the same section naming convention as Linux above it, where it says x86 for both i686 and x86-64.

ViralBShah:
I guess technically, the freebsd and mac should both be blank in the 32-bit sections. Not a big deal though. Just mentioned it since I noticed it.

PR #812, opened 2018-07-09
Benchmark algorithm descriptions

johnfgibson:
Add brief descriptions of microbenchmark algorithms to benchmarks page, in order to make the what, how and why of each algorithm perfectly clear. 

I also put in links from each descriptions to the corresponding function definitions in the current perf.jl blob on github. The links will have to be updated manually if the Julia benchmark kernels change significantly, but I don't expect they will, and even so it seems better than linking to meandering line numbers on master or cut-and-pasting code snippets into the benchmarks page.

### DIFF
--- a/benchmarks/index.md
+++ b/benchmarks/index.md
@@ -10,23 +10,6 @@ performance on a range of common code patterns, such as function
 calls, string parsing, sorting, numerical loops, random number
 generation, recursion, and array operations.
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Needs to be done on the new website.

https://github.com/JuliaLang/www.julialang.org/issues/31

PR #818, opened 2018-07-30
Update index.md

manuel1957:
there is no debugger include in juliapro either free or profesionnal


ViralBShah:
Should come back post 1.0 but good to remove for now.

